PR #4, opened 2024-03-14
feat: Add Milvus vector database

leehuwuj:


helpers/typescript.ts
```diff
@@ -231,6 +240,13 @@ export const installTSTemplate = async ({
     };
   }
 
+  if (vectorDb === "milvus") {
+    packageJson.dependencies = {
+      ...packageJson.dependencies,
+      "@zilliz/milvus2-sdk-node": "^2.3.5",
+    };
+  }
```
marcusschiesser, line 248: i think we don't need this, as`llamaindex` already has the dependencies

helpers/env-variables.ts
```diff
@@ -71,6 +71,27 @@ const getVectorDBEnvs = (vectorDb: TemplateVectorDB) => {
           name: "PINECONE_INDEX_NAME",
         },
       ];
+    case "milvus":
+      return [
```
marcusschiesser, line 75: also generated for TS?

templates/components/vectordbs/typescript/milvus/generate.mjs
```diff
@@ -0,0 +1,44 @@
...
+async function loadAndIndex() {
+  // load objects from storage and convert them into LlamaIndex Document objects
+  const documents = await new SimpleDirectoryReader().loadData({
+    directoryPath: STORAGE_DIR,
+  });
+
+  // Connect to Milvus
+  const milvusClient = getMilvusClient();
+  const vectorStore = new MilvusVectorStore({ milvusClient });
+  await vectorStore.connect();
```
marcusschiesser, line 28: remove after fixing in the store

marcusschiesser (approved):

PR #5, opened 2024-03-14
feat: support agent in typescript templates

thucpn:
Done:
- move chat engine from vectordbs to engine folder
- update route & stream handler for nextjs and express templates
- add questions to choose install agents for typescript templates
- add WikipediaTool in packages/core (PR: https://github.com/run-llama/LlamaIndexTS/pull/648)
- create ToolFactory and load WikipediaTool to agent chat engine


marcusschiesser (commented):
looking good. we should also add an e2e test for selecting the agent

templates/components/engines/typescript/agent/tools.ts
```diff
@@ -1 +1,37 @@
-// Load tools here
\ No newline at end of file
+import { BaseTool } from "llamaindex";
+import config from "./tool_config.json";
+
+enum ExternalTool {
+  GoogleSearch = "google.GoogleSearchToolSpec",
```
marcusschiesser, line 5: `GoogleSearch` can be removed for now

templates/components/engines/typescript/agent/tools.ts
```diff
@@ -1 +1,37 @@
...
+    }
+
+    // TODO: Implement GoogleSearchToolSpec
+
+    throw new Error(
+      `Sorry! Tool ${key} is not supported yet. Options: ${options}`,
+    );
+  }
+
+  public static async list(): Promise<BaseTool[]> {
```
marcusschiesser, line 29: 1. separate config from the factory 
2. How about moving the factory to `LlamaIndexTS` (might be useful for other use cases)? 

```suggestion
  public static async createTools(config: ToolConfig): Promise<BaseTool[]> {
```

templates/components/engines/typescript/agent/tools.ts
marcusschiesser: Is `questions.ts` only asking for defined tools?

helpers/tools.ts
```diff
@@ -37,6 +38,7 @@ export const supportedTools: Tool[] = [
         version: "0.1.2",
       },
     ],
+    tsSupport: true,
```
marcusschiesser, line 41: we might later also have tools that only work in TS, how about
```suggestion
    supportedLanguages: ["python","typescript"],
```

marcusschiesser (approved):

PR #10, opened 2024-03-19
Feat: Add multiple data sources

leehuwuj:


helpers/env-variables.ts
```diff
@@ -99,26 +99,28 @@ const getVectorDBEnvs = (vectorDb: TemplateVectorDB) => {
...
-        },
-      ];
-    default:
-      return [];
+const getDataSourceEnvs = (dataSources: TemplateDataSource[]) => {
+  const envs = [];
+  for (const source of dataSources) {
+    switch (source.type) {
+      case "web":
+        envs.push(
```
marcusschiesser, line 107: might we get duplicates with that approach?

templates/components/vectordbs/python/none/generate.py
```diff
@@ -16,15 +16,17 @@
...
-        documents,
-    )
-    # store it for later
-    index.storage_context.persist(STORAGE_DIR)
-    logger.info(f"Finished creating new index. Stored in {STORAGE_DIR}")
+    storage_context = StorageContext.from_defaults()
+
+    docs = []
+
+    for doc in get_documents():
```
marcusschiesser, line 23: why not returning all documents here (instead of the iterator)?

templates/components/loaders/python/loader.py
```diff
@@ -0,0 +1,12 @@
+import os
+import importlib
+
+
+def get_documents():
```
marcusschiesser, line 5: that's nice

templates/components/vectordbs/python/none/generate.py
leehuwuj: ah, we may have more actions for each returned document so i implemented the iterator here, will convert to the document list if the iterator is not needed.

PR #15, opened 2024-03-22
feat: Add multi files/folders selection

leehuwuj:
Note: Multi folder selection is not supported in Windows


templates/components/data/TOS.pdf
marcusschiesser, line 1: Why copy another example?

leehuwuj: We'll support multi documents, so it'll be great if we add another example, don't you think it is?

marcusschiesser: Wouldn't do that as it slows the startup of the generated app and increases the costs Generating the embeddings.

marcusschiesser (approved):

PR #18, opened 2024-03-25
feat: Add multiple URLs reader

leehuwuj:


questions.ts
```diff
@@ -755,35 +755,56 @@ export const askQuestions = async (
...
+        },
+        {
+          type: "text",
+          name: "prefix",
+          message:
+            "Please provide prefix for the URLs you would like to crawl: ",
+          initial: "https://www.llamaindex.ai/enterprise",
+          validate: validateUrl,
+        },
+        {
```
marcusschiesser, line 790: why do we need two parameters for the URL?

helpers/env-variables.ts
```diff
@@ -105,25 +104,6 @@ const getDataSourceEnvs = (
   llamaCloudKey?: string,
 ) => {
   switch (dataSource.type) {
```
marcusschiesser, line 106: as we're moving the datasource config now to json with this PR, it's confusing to keep this code. 
how about generating `LLAMA_CLOUD_API_KEY` all the time? it simplifies the code and should not hurt

templates/components/loaders/python/web/loader.py
```diff
@@ -1,13 +1,32 @@
...
 
 
-def get_documents():
-    # Initialize the scraper with a prefix URL and maximum depth
-    scraper = WholeSiteReader(
-        prefix=os.environ.get("URL_PREFIX"), max_depth=int(os.environ.get("MAX_DEPTH"))
-    )
-    # Start scraping from a base URL
-    documents = scraper.load_data(base_url=os.environ.get("BASE_URL"))
+class LoaderConfig(BaseModel):
```
marcusschiesser, line 7:
```suggestion
class WebLoaderConfig(BaseModel):
```

helpers/python.ts
```diff
@@ -268,6 +269,29 @@ export const installPythonTemplate = async ({
...
+    if (dataSource?.type === "web") {
+      const config = dataSource.config as WebSourceConfig[];
+      const webLoaderConfig = config.map((c) => {
+        return {
+          base_url: c.baseUrl,
+          prefix: c.prefix || c.baseUrl,
+          depth: c.depth || 1,
+        };
+      });
+      const loaderConfigPath = path.join(root, "loaders.json");
```
marcusschiesser, line 283: better move it to config and also move the tools config
```suggestion
      const loaderConfigPath = path.join(root, "config/loaders.json");
```

marcusschiesser: this means we also need one two sentences in the README that that the loaders and tools can be configured there

templates/components/loaders/python/web/loader.py
```diff
@@ -1,13 +1,32 @@
...
+    max_depth: int = Field(default=1, ge=0)
+
+
+def load_configs():
+    with open("loaders.json") as f:
+        configs = json.load(f)
+    web_config = configs.get("web", None)
+    if web_config is None:
+        raise ValueError("No web config found in loaders.json")
+    print(web_config)
```
marcusschiesser, line 19: better log if needed

templates/components/loaders/python/web/loader.py
```diff
@@ -1,13 +1,32 @@
...
+    return [LoaderConfig(**config) for config in web_config]
 
+
+def get_documents():
+    web_config = load_configs()
```
marcusschiesser, line 24: not sure if this is better, but you could consider separating the load and parse concern, e.g.:
```suggestion
    config = parse_config(load_config())
```

templates/components/loaders/python/web/loader.py
```diff
@@ -1,13 +1,32 @@
...
+    documents = []
+    for config in web_config:
```
marcusschiesser, line 26: we could say that we have one config but multiple config entries (or items):
```suggestion
    for entry in config:
```

questions.ts
leehuwuj: It allows users to config more crawler parameters, i ignored the prefix to simplify the cli questions before. If you think it's confusing the user, then we can remove it and consider adding more guidelines once we move to yaml config.

helpers/env-variables.ts
leehuwuj: Ah yes, i removed the env code, looks like I added it back when rebasing.

questions.ts
marcusschiesser: yes i think it's better. we already have to many parameters for the initial generation. 
Initial generation should be easy (so user are not leaving it) and then later they can configure and tweak more

templates/components/loaders/python/web/loader.py
leehuwuj: It's fine for me for now, of course we can update it once we update yaml config or adding more configuration for loaders

templates/components/loaders/python/web/loader.py
```diff
@@ -1,13 +1,31 @@
...
+    with open("config/loaders.json") as f:
+        configs = json.load(f)
+    web_config = configs.get("web", None)
+    if web_config is None:
+        raise ValueError("No web config found in loaders.json")
+    return [WebLoaderConfig(**config) for config in web_config]
 
+
+def get_documents():
+    web_config = load_configs()
```
marcusschiesser, line 23: ok, i'll merge this now. in the next PR the naming should be more consistent, like:
```suggestion
    config = load_config()
```
or
```suggestion
    web_config = load_web_config()
```

marcusschiesser (approved):

PR #19, opened 2024-03-25
feat: Add multiple data sources

leehuwuj:


questions.ts
```diff
@@ -625,182 +642,229 @@ export const askQuestions = async (
...
-            type: "toggle",
-            name: "useLlamaParse",
-            message:
-              "Would you like to use LlamaParse (improved parser for RAG - requires API key)?",
-            initial: true,
-            active: "yes",
-            inactive: "no",
-          },
-          handlers,
+      if (!program.useLlamaParse && program.engine === "context") {
```
marcusschiesser, line 830: do we need this condition?

questions.ts
```diff
@@ -625,182 +642,229 @@ export const askQuestions = async (
...
+        const askingLlamaParse = program.dataSources.some(
+          (ds) =>
+            ds.type === "folder" ||
+            (ds.type === "file" &&
+              (ds.config as FileSourceConfig).paths?.some(
+                (p) => path.extname(p) === ".pdf",
+              )),
```
marcusschiesser, line 837: i think we can simplify this and always ask for llamaparse if we select file or folder

questions.ts
```diff
@@ -625,182 +642,229 @@ export const askQuestions = async (
...
+            ),
+            initial: 0,
+          },
+          handlers,
+        );
+
+        // Asking for data source config
+        // Select None data source, No need to config and asking for another data source
+        if (selectedSource === "none") {
+          if (selectedSource.length === 0) {
```
marcusschiesser, line 684: is this correct? once `selectedSource` is a string and once an array

templates/components/loaders/python/loader.py
```diff
@@ -0,0 +1,13 @@
...
+import importlib
+
+
+def get_documents():
+    # For each file in .loaders, import the module and call the get_documents function
+    documents = []
+    for loader in os.listdir(os.path.join(os.path.dirname(__file__), "loaders")):
+        if loader.endswith(".py"):
+            loader = loader[:-3]
+            module = importlib.import_module(f"app.engine.loaders.{loader}")
```
marcusschiesser, line 11: 1. so we just import one file per loader? this means the loader files must have different names - we should probably write a comment
2. do we copy this loader only if we have multiple datasources?

leehuwuj: 1. Yes, we only copy one file per loader. 
There is another approach which is more straightforward to the users is specifying all the load_documents from each loader in a __init__.py file in the loaders folder, but this approach requires generate init file dynamically so i did not implement that way.
2. Yes, we still copy this loader if there is only one data source.

questions.ts
leehuwuj: Thanks, great catch! It should `program.dataSources.length` but i updated the code to make it easier to read.

questions.ts
leehuwuj: removed, it is redundant. Thanks!

questions.ts
```diff
@@ -625,122 +641,210 @@ export const askQuestions = async (
...
+
+        // No need to ask for another data source if user selected example data
+        if (selectedSource === "exampleFile") {
+          break;
         }
       }
+
+      if (
+        program.dataSources.length === 0 ||
+        program.dataSources[0].type === "none"
```
marcusschiesser, line 795: can we get rid of this none type by keeping the array empty (means no datasource?)

questions.ts
```diff
@@ -625,122 +641,210 @@ export const askQuestions = async (
...
     } else if (program.engine === "simple") {
-      program.dataSource = {
-        type: "none",
-        config: {},
-      };
+      program.dataSources = [
+        {
+          type: "none",
+          config: {},
+        },
```
marcusschiesser, line 816: this would then be an empty array

questions.ts
```diff
@@ -625,122 +641,210 @@ export const askQuestions = async (
...
-            program.dataSource = { type: "folder", config: {} };
-            break;
-          case "localFile":
-            program.engine = "context";
-            program.dataSource = {
-              type: "file",
-              config: {
-                paths: await selectLocalContextData("file"),
+      program.dataSources = [];
+      const numberDataSources = program.framework === "fastapi" ? 2 : 1;
```
marcusschiesser, line 661: would that logic be easier if we keep asking this question?
then we could also remove the while(true) for the URLs

helpers/python.ts
```diff
@@ -256,51 +257,62 @@ export const installPythonTemplate = async ({
...
-      }
-      await copy("**", enginePath, {
+      // Copy loader.py file to enginePath
+      await copy("loader.py", enginePath, {
         parents: true,
-        cwd: path.join(compPath, "loaders", "python", loaderFolder),
+        cwd: path.join(compPath, "loaders", "python"),
       });
+
+      for (const dataSource of dataSources) {
```
marcusschiesser, line 270: if we have multiple file or web datasources, we copy the file here multiples times, right?

templates/components/loaders/python/loader.py
marcusschiesser: if we always copy this loader, how about we always copy the file (or llamaparse) and the web loader too?
we could then also make this import not dynamic but explicit

helpers/typescript.ts
```diff
@@ -175,7 +175,7 @@ export const installTSTemplate = async ({
 
     // copy loader component
     const dataSourceType = dataSources[0]?.type;
```
marcusschiesser, line 177: this doesn't work on an empty array

leehuwuj: already have a question mark there so it should be ok

marcusschiesser: you're right - wrote to much java code in my life where it would throw an exception :D

templates/components/loaders/python/__init__.py
```diff
@@ -0,0 +1,33 @@
+import os
```
marcusschiesser, line 1: loaders looking great now!

questions.ts
```diff
@@ -707,80 +699,43 @@ export const askQuestions = async (
...
-          }
-          program.dataSources.push(dataSource);
+            },
+          });
         }
 
-        // No need to ask for another data source if user selected example data
-        if (selectedSource === "exampleFile") {
+        if (
+          program.framework !== "fastapi" ||
```
marcusschiesser, line 736: we probably can leave this out to simplify?

package.json
```diff
@@ -68,7 +69,8 @@
     "prettier-plugin-organize-imports": "^3.2.4",
     "typescript": "^5.3.3",
     "eslint-config-prettier": "^8.10.0",
-    "ora": "^8.0.1"
+    "ora": "^8.0.1",
+    "fs-extra": "11.2.0"
```
marcusschiesser, line 73: where do we need this?

marcusschiesser: ah sorry, found it

questions.ts
```diff
@@ -620,127 +630,149 @@ export const askQuestions = async (
...
+
+      if (program.dataSources.length === 0) {
+        program.engine = "simple";
+      } else {
+        program.engine = "context";
+      }
     }
-  } else if (!program.dataSource) {
+  } else if (!program.dataSources) {
     // Handle a case when engine is specified but dataSource is not
```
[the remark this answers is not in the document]
leehuwuj: I added this before for the case user use `--engine context` argument but did not specify the dataSource to use, ex(`--files`). It also applies for e2e test where it doesn't have a pdf file then it should use the example pdf file.

marcusschiesser (approved):

PR #20, opened 2024-03-25
refactor: make non-streaming chat as /chat/request route

thucpn:


templates/types/streaming/fastapi/app/api/routers/chat.py
```diff
@@ -18,20 +18,18 @@ class _ChatData(BaseModel):
...
-    data: _ChatData,
-    chat_engine: BaseChatEngine = Depends(get_chat_engine),
-):
-    # check preconditions and get last message
+class _Result(BaseModel):
+    result: _Message
+
+
+async def preprocess_request(data: _ChatData) -> tuple:
     if len(data.messages) == 0:
```
marcusschiesser, line 26:
```suggestion
    # check preconditions and get last message
    if len(data.messages) == 0:
```

templates/types/streaming/fastapi/app/api/routers/chat.py
```diff
@@ -18,20 +18,18 @@ class _ChatData(BaseModel):
...
-    request: Request,
-    data: _ChatData,
-    chat_engine: BaseChatEngine = Depends(get_chat_engine),
-):
-    # check preconditions and get last message
+class _Result(BaseModel):
+    result: _Message
+
+
+async def preprocess_request(data: _ChatData) -> tuple:
```
marcusschiesser, line 25:
```suggestion
from typing import List, Tuple

# parses _ChatData to be used by LlamaIndex chat
async def parse_chat_data(data: _ChatData) -> Tuple[str, List[ChatMessage]]:
```

templates/types/streaming/fastapi/app/api/routers/chat.py
```diff
@@ -44,16 +42,36 @@ async def chat(
         )
         for m in data.messages
     ]
+    return last_message, messages
```
marcusschiesser, line 45: Here we're a bit in a pickle as last_message is a different message type than messages, to be consistent we could either return:

```suggestion
    return last_message.content, messages
```
or
```suggestion
    return ChatMessage(role=last_message.role,
            content=last_message.content), messages
```

I think the first one is easier to understand

templates/types/streaming/fastapi/app/api/routers/chat.py
```diff
@@ -44,16 +42,36 @@ async def chat(
...
 
-    # query chat engine
-    response = await chat_engine.astream_chat(lastMessage.content, messages)
 
-    # stream response
+@r.post("")
```
marcusschiesser, line 48:
```suggestion
# streaming endpoint - delete if not needed
@r.post("")
```

templates/types/streaming/fastapi/app/api/routers/chat.py
```diff
@@ -44,16 +42,36 @@ async def chat(
...
         async for token in response.async_response_gen():
-            # If client closes connection, stop sending events
             if await request.is_disconnected():
                 break
             yield token
 
     return StreamingResponse(event_generator(), media_type="text/plain")
+
+
+@r.post("/request")
```
marcusschiesser, line 67:
```suggestion
# non-streaming endpoint - delete if not needed
@r.post("/request")
```

marcusschiesser: @thucpn i think in the FastAPI code, it's easy to just delete the function endpoint, so we can leave it in one file for simplicity

templates/types/streaming/express/src/routes/chat.route.ts
```diff
@@ -1,8 +1,10 @@
 import express from "express";
+import { chatRequest } from "../controllers/chat-request.controller";
 import { chat } from "../controllers/chat.controller";
```
marcusschiesser, line 3: @thucpn did you delete that?

marcusschiesser: ah no, sorry

marcusschiesser (approved):
@thucpn thanks looks great

PR #22, opened 2024-03-27
fix: remove redundant template engine variable and fixed llamaparse flag

marcusschiesser:


index.ts
```diff
@@ -71,13 +71,6 @@ const program = new Commander.Command(packageJson.name)
     `
 
   Select a template to bootstrap the application with.
-`,
```
leehuwuj, line 74: If we remove `engine` so there is no flag for simple chat, then the cli always asks for the data source. I think we can add a `--no-data-source` option for that case

index.ts
```diff
@@ -199,7 +199,7 @@ if (process.argv.includes("--tools")) {
   }
 }
 if (process.argv.includes("--no-llama-parse")) {
-  program.llamaParse = false;
```
leehuwuj, line 202: great catch!

index.ts
marcusschiesser: ah that's why my tests failed! thanks

PR #26, opened 2024-03-27
feat: Use yaml format instead of json for loaders and tools config

leehuwuj:


create-app.ts
```diff
@@ -130,8 +130,8 @@ export async function createApp({
     console.log(
       yellow(
         `You have selected tools that require configuration. Please configure them in the ${terminalLink(
-          "config/tools.json",
-          `file://${root}/config/tools.json`,
+          "config/tools.yaml",
```
marcusschiesser, line 133: Do we also show this for typescript?

marcusschiesser: This is in the shared code and python specific. We either have to move it to the python code or better language agnostic

leehuwuj: Yes, it's python specific but putting it here will help user have a notice about that.

PR #28, opened 2024-03-29
feat: Add database data source

leehuwuj:


questions.ts
```diff
@@ -629,52 +633,88 @@ export const askQuestions = async (
...
-                }
-                const urlObj = new URL(value);
-                if (
-                  urlObj.protocol !== "https:" &&
-                  urlObj.protocol !== "http:"
-                ) {
-                  return `URL=${value} has invalid protocol, only allow http or https`;
-                }
-                return true;
+          case "db": {
```
leehuwuj, line 688: We'll allow multiple database sources, each database source can have multiple queries, but we only ask one query in the CLI, user can reconfigure it later in the loaders.yaml file.

.changeset/seven-zebras-allow.md
```diff
@@ -0,0 +1,5 @@
+---
+"create-llama": patch
+---
+
+Add database data source
```
marcusschiesser, line 5:
```suggestion
Use databases as data source
```

questions.ts
marcusschiesser: that's great. keep the UI simple as possible

templates/components/loaders/python/db.py
```diff
@@ -0,0 +1,29 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class DBLoaderConfig(BaseModel):
+    uri: str
+    queries: List[str]
+
+
+def get_db_documents(raw_configs: list[dict]):
```
marcusschiesser, line 15: why not separating parse the config and using the config?
```suggestion
def get_db_documents(raw_configs: list[DBLoaderConfig]):
```

helpers/python.ts
```diff
@@ -65,17 +66,34 @@ const getAdditionalDependencies = (
...
+      dependencies.push({
+        name: "llama-index-readers-database",
+        version: "^0.1.3",
+      });
+      dependencies.push({
+        name: "pymysql",
+        version: "^1.1.0",
+        extras: ["rsa"],
+      });
+      dependencies.push({
```
leehuwuj, line 92: This will add both pymysql and psycopg when use selected the db source but it's fine for me. We can separate the depepdencies based on the schema later on if we have other databases.
Using psycopg (psycopg 3) driver is the best option for now.

marcusschiesser: Weird that it also add pymysql. To make this more explicit how about adding the  pymysql version manually that is added automatically?

marcusschiesser (approved):

PR #33, opened 2024-04-02
Allow tools without datasource

marcusschiesser:
Simplify code generation

### DIFF
--- a/helpers/index.ts
+++ b/helpers/index.ts
@@ -119,12 +119,17 @@ export const installTemplate = async (
 
   if (props.framework === "fastapi") {
     await installPythonTemplate(props);
+    // write loaders configuration (currently Python only)
+    await writeLoadersConfig(
+      props.root,
+      props.dataSources,
+      props.useLlamaParse,
+    );
   } else {
     await installTSTemplate(props);
   }
 
-  // write configuration files
-  await writeLoadersConfig(props.root, props.dataSources, props.useLlamaParse);
+  // write tools configuration
   await writeToolsConfig(
     props.root,
     props.tools,
--- a/helpers/index.ts
+++ b/helpers/index.ts
@@ -4,12 +4,14 @@ import path from "path";
 import { cyan } from "picocolors";
 
 import fsExtra from "fs-extra";
+import { writeLoadersConfig } from "./datasources";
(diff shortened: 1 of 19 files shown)


templates/components/engines/typescript/agent/chat.ts
```diff
@@ -1,26 +1,44 @@
...
+      }),
+    );
+  }
 
-  const externalTools = await ToolFactory.createTools(config);
+  try {
+    // add tools from config file if it exists
+    const config = JSON.parse(
+      await fs.readFile(path.join("config", "tools.json"), "utf8"),
+    );
```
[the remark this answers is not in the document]
marcusschiesser: to support that this file doesn't exist (no tools selected) - I simplified the generator that we either always use the ContextChatEngine or the AgentChatEngine

templates/components/engines/python/agent/__init__.py
```diff
@@ -11,11 +11,12 @@ def get_chat_engine():
     top_k = os.getenv("TOP_K", "3")
     tools = []
 
-    # Add query tool
+    # Add query tool if index exists
```
leehuwuj, line 14: Does this mode work if the user does not have an index and no tools config as well?

templates/components/engines/typescript/agent/chat.ts
```diff
@@ -1,26 +1,44 @@
...
+    tools = tools.concat(await ToolFactory.createTools(config));
```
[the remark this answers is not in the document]
marcusschiesser: good idea. i just kept it as as is it's probably a bit better to read

templates/components/engines/python/agent/__init__.py
marcusschiesser: yes, tested this

helpers/index.ts
```diff
@@ -121,6 +123,14 @@ export const installTemplate = async (
     await installTSTemplate(props);
   }
 
+  // write configuration files
+  await writeLoadersConfig(props.root, props.dataSources, props.useLlamaParse);
```
leehuwuj, line 127: TS does not support loaders right now, right? I think we should only generate loaders config for python only for now

marcusschiesser: @leehuwuj thanks good idea!

PR #37, opened 2024-04-05
fix: remove empty object at start and end of streaming

thucpn:
## Summary by CodeRabbit

- **Bug Fixes**
	- Improved message handling by eliminating the sending of empty image responses for both user and assistant messages.
	- Modified logic in `llamaindex-stream.ts` to remove the behavior of sending empty image responses for messages.
	- Updated conditional checks in `llamaindex-stream.ts` based on `process.env.MODEL` value for message response generation.

### DIFF
--- a/templates/types/streaming/express/src/controllers/llamaindex-stream.ts
+++ b/templates/types/streaming/express/src/controllers/llamaindex-stream.ts
@@ -30,15 +30,12 @@ function createParser(
           },
         };
         data.append(message);
-      } else {
-        data.append({}); // send an empty image response for the user's message
       }
     },
     async pull(controller): Promise<void> {
       const { value, done } = await it.next();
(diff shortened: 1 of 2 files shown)


templates/types/streaming/express/src/controllers/llamaindex-stream.ts
```diff
@@ -30,15 +30,12 @@ function createParser(
           },
         };
         data.append(message);
-      } else {
-        data.append({}); // send an empty image response for the user's message
```
marcusschiesser, line 34: @thucpn I remember I added this to ensure that the client always receives a data object for each message. That way it's easy to map the message with the data object. 
I don't think images will work if we remove this

marcusschiesser:
won't fix

PR #70, opened 2024-05-01
chore: bump to llamaindex 0.3

himself65:
We don't need modify webpack anymore

## Summary by CodeRabbit

- **Dependency Updates**
  - Updated the `llamaindex` package from version `0.2.10` to `0.3.1` to introduce potential new features and improvements.

- **Configuration Changes**
  - Removed specific dependency resolution and external module handling configurations in the NextJS webpack setup.
  - Eliminated the `aws4` fallback configuration in the NextJS webpack setup for streaming types.

### DIFF
--- a/templates/types/streaming/nextjs/webpack.config.o11y.mjs
+++ b/templates/types/streaming/nextjs/webpack.config.o11y.mjs
@@ -1,7 +1,4 @@
 export default function webpack(config, isServer) {
-  config.resolve.fallback = {
-    aws4: false,
-  };
   config.module.rules.push({
     test: /\.node$/,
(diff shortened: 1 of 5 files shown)


templates/types/streaming/nextjs/webpack.config.o11y.mjs
```diff
@@ -1,7 +1,4 @@
 export default function webpack(config, isServer) {
-  config.resolve.fallback = {
```
marcusschiesser, line 2: @himself65 as far as i remember, this is necessary when using MongoDB vector store, why did you remove this?

PR #71, opened 2024-05-02
Add persist index store for Simple Vector Store

leehuwuj:
To fix the issue of not retrieving any nodes.

## Summary by CodeRabbit

- **New Features**
	- Enhanced data persistence logic to support different types of vector stores.
	- Improved index loading conditions for more efficient data retrieval.

- **Bug Fixes**
	- Standardized namespace usage in vector storage to "default" for consistency.

- **Refactor**
	- Updated parameter name from `dimension` to `dimensions` in OpenAI initialization settings for clarity.

### DIFF
--- a/templates/components/vectordbs/python/none/vectordb.py
+++ b/templates/components/vectordbs/python/none/vectordb.py
@@ -6,10 +6,11 @@ from app.constants import STORAGE_DIR
 
 def get_vector_store():
     if not os.path.exists(STORAGE_DIR):
+        # Vector store hasn't been persisted before, create a new one
         vector_store = SimpleVectorStore()
     else:
+        # Vector store has already been persisted before at STORAGE_DIR - load it
         vector_store = SimpleVectorStore.from_persist_dir(
             STORAGE_DIR, namespace="default"
         )
-    vector_store.stores_text = True
     return vector_store
--- a/templates/components/vectordbs/python/none/vectordb.py
+++ b/templates/components/vectordbs/python/none/vectordb.py
@@ -8,6 +8,8 @@ def get_vector_store():
     if not os.path.exists(STORAGE_DIR):
         vector_store = SimpleVectorStore()
     else:
-        vector_store = SimpleVectorStore.from_persist_dir(STORAGE_DIR)
+        vector_store = SimpleVectorStore.from_persist_dir(
+            STORAGE_DIR, namespace="default"
+        )
     vector_store.stores_text = True
     return vector_store
--- a/templates/types/streaming/fastapi/app/engine/generate.py
+++ b/templates/types/streaming/fastapi/app/engine/generate.py
@@ -70,10 +70,9 @@ def generate_datasource():
             ),
             store_nodes_override=True,
         )
-        # SimpleDocumentStore keeps the data in memory only - needs to be persisted explicitly
         index.storage_context.persist(STORAGE_DIR)
     else:
-        # Persist the docstore to apply ingestion strategy
+        # SimpleDocumentStore keeps the data in memory only - needs to be persisted explicitly
         docstore.persist(os.path.join(STORAGE_DIR, DEFAULT_PERSIST_FNAME))
 
     logger.info("Finished creating new index.")
--- a/templates/types/streaming/fastapi/app/engine/generate.py
+++ b/templates/types/streaming/fastapi/app/engine/generate.py
@@ -72,7 +72,7 @@ def persist_storage(docstore, vector_store, nodes):
 
 def generate_datasource():
     init_settings()
-    logger.info("Indexing the data")
+    logger.info("Generate index for the provided data")
 
     # Get the stores and documents or create new ones
     documents = get_documents()
@@ -89,7 +89,7 @@ def generate_datasource():
     # Build the index and persist storage
     persist_storage(docstore, vector_store, nodes)
 
-    logger.info("Finished the indexing")
+    logger.info("Finished generating the index")
 
 
 if __name__ == "__main__":
--- a/templates/types/streaming/fastapi/app/engine/generate.py
+++ b/templates/types/streaming/fastapi/app/engine/generate.py
@@ -9,6 +9,9 @@ from llama_index.core.ingestion import IngestionPipeline
 from llama_index.core.node_parser import SentenceSplitter
 from llama_index.core.vector_stores import SimpleVectorStore
 from llama_index.core.storage.docstore import SimpleDocumentStore
+from llama_index.core.storage.index_store import SimpleIndexStore
+from llama_index.core.storage import StorageContext
+from llama_index.core import VectorStoreIndex
 from app.constants import STORAGE_DIR
 from app.settings import init_settings
 from app.engine.loaders import get_documents
@@ -54,14 +57,23 @@ def generate_datasource():
     ingestion_pipeline.vector_store = vector_store
 
     # Run the ingestion pipeline and store the results
-    ingestion_pipeline.run(show_progress=True, documents=documents)
+    nodes = ingestion_pipeline.run(show_progress=True, documents=documents)
 
     # Default vector store only keeps data in memory, so we need to persist it
     # Can remove if using a different vector store
     if isinstance(vector_store, SimpleVectorStore):
-        vector_store.persist(os.path.join(STORAGE_DIR, "vector_store.json"))
-    # Persist the docstore to apply ingestion strategy
-    docstore.persist(os.path.join(STORAGE_DIR, "docstore.json"))
+        index = VectorStoreIndex(
+            nodes=nodes,
+            storage_context=StorageContext.from_defaults(
+                docstore=docstore,
+                vector_store=vector_store,
+            ),
+            store_nodes_override=True,
+        )
+        index.storage_context.persist(STORAGE_DIR)
+    else:
+        # Persist the docstore to apply ingestion strategy
+        docstore.docstore.persist(os.path.join(STORAGE_DIR, "docstore.json"))
 
     logger.info("Finished creating new index.")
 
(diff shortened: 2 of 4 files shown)


templates/components/vectordbs/python/none/vectordb.py
```diff
@@ -8,6 +8,8 @@ def get_vector_store():
     if not os.path.exists(STORAGE_DIR):
         vector_store = SimpleVectorStore()
     else:
-        vector_store = SimpleVectorStore.from_persist_dir(STORAGE_DIR)
+        vector_store = SimpleVectorStore.from_persist_dir(
+            STORAGE_DIR, namespace="default"
```
marcusschiesser, line 12: 1. isn't `default` the default namespace - do we need to specify it?
2. can't we just use:
`vector_store = SimpleVectorStore.from_persist_dir(
            STORAGE_DIR)` also if the path `STORAG_DIR` doesn't exit?

templates/types/streaming/fastapi/app/engine/generate.py
```diff
@@ -54,14 +57,23 @@ def generate_datasource():
     ingestion_pipeline.vector_store = vector_store
 
     # Run the ingestion pipeline and store the results
-    ingestion_pipeline.run(show_progress=True, documents=documents)
+    nodes = ingestion_pipeline.run(show_progress=True, documents=documents)
 
     # Default vector store only keeps data in memory, so we need to persist it
     # Can remove if using a different vector store
```
marcusschiesser, line 63:
```suggestion
    # Can be removed if using a different vector store
```

templates/types/streaming/fastapi/app/engine/generate.py
```diff
@@ -54,14 +57,23 @@ def generate_datasource():
...
+            storage_context=StorageContext.from_defaults(
+                docstore=docstore,
+                vector_store=vector_store,
+            ),
+            store_nodes_override=True,
+        )
+        index.storage_context.persist(STORAGE_DIR)
+    else:
+        # Persist the docstore to apply ingestion strategy
+        docstore.persist(os.path.join(STORAGE_DIR, "docstore.json"))
```
marcusschiesser, line 76: any way to do this without specifying `docstore.json`? shouldn't it be the default? don't we need

marcusschiesser, line 76: i guess we only need to do this if we have a file-based docstore (which we're having at the moment anyhow only):
```suggestion
        # SimpleDocumentStore keeps the data in memory only - needs to be persisted explicitly 
        if isinstance(docstore, SimpleDocumentStore):
          docstore.persist(os.path.join(STORAGE_DIR, "docstore.json"))
```

templates/types/streaming/fastapi/app/engine/generate.py
```diff
@@ -54,14 +57,23 @@ def generate_datasource():
     ingestion_pipeline.vector_store = vector_store
 
     # Run the ingestion pipeline and store the results
-    ingestion_pipeline.run(show_progress=True, documents=documents)
+    nodes = ingestion_pipeline.run(show_progress=True, documents=documents)
 
     # Default vector store only keeps data in memory, so we need to persist it
```
marcusschiesser, line 62:
```suggestion
    # SimpleVectorStore keeps the data in memory only - needs to be persisted explicitly
```

templates/types/streaming/fastapi/app/engine/generate.py
```diff
@@ -54,14 +57,23 @@ def generate_datasource():
...
     # Can remove if using a different vector store
     if isinstance(vector_store, SimpleVectorStore):
-        vector_store.persist(os.path.join(STORAGE_DIR, "vector_store.json"))
-    # Persist the docstore to apply ingestion strategy
-    docstore.persist(os.path.join(STORAGE_DIR, "docstore.json"))
+        index = VectorStoreIndex(
+            nodes=nodes,
+            storage_context=StorageContext.from_defaults(
```
marcusschiesser, line 67: `StorageContext` has a `persist` function that persists vec, index and doc store, can't we use it instead?

templates/types/streaming/fastapi/app/engine/generate.py
leehuwuj: It's a path. Because we have custom storage config so we must construct the path instead of using default. The "docstore.json" is already default value, but let me make it better by importing the default variable instead.

templates/components/vectordbs/python/none/vectordb.py
leehuwuj: Oh, I see. It can be a bit confusing. Basically, the llama_index stores the index with a prefix (namespace). However, when loading it, if we don't specify the namespace, it won't use the default one.

Great catch! we should add a logic to assert the missing STORAGE_DIR if it doesn't exit.

templates/types/streaming/fastapi/app/engine/generate.py
leehuwuj: Using storage_context is fine, but it'll also generate other files like vector store, index store,.. which are redundant for other vector store. 
Explicitly persist for docstore is not a bad idea i think, it'll clearly show to the user that the ingestion pipeline for vector store only need the docstore.

templates/components/vectordbs/python/none/vectordb.py
marcusschiesser: About 2. I mean why do we need the case with SimpleVectorStore()

templates/components/vectordbs/python/none/vectordb.py
```diff
@@ -8,6 +8,8 @@ def get_vector_store():
...
+        )
     vector_store.stores_text = True
```
marcusschiesser, line 14: Stores_text = true is needed? If so a comment would be good

templates/components/vectordbs/python/none/vectordb.py
leehuwuj: Ah, understand. The SimpleVectorStore() case is for generating step where it initializes a new simple vector store instance. Once we already have the vector store (persisted previously) then we should initialize the vector store with persisted data.

templates/types/streaming/fastapi/app/engine/generate.py
leehuwuj: For now, we're using ingestion pipeline with a strategy which is required a docstore, and the file base docstore is default option. We also need it for all the vector store too.

templates/types/streaming/fastapi/app/engine/generate.py
leehuwuj: Just answered above, please check: https://github.com/run-llama/create-llama/pull/71#discussion_r1587025184

templates/components/vectordbs/python/none/vectordb.py
leehuwuj: Previously, i used VectorStoreIndex.from_vector_store() so the vector_store must enable the stores_text (but it's not able to load the index). But now i'm using the load_index_from_storage so the stores_text could be redundant, let me check.

templates/components/vectordbs/python/none/vectordb.py
marcusschiesser: got it i'll add some comments

templates/components/vectordbs/python/none/vectordb.py
```diff
@@ -8,6 +8,8 @@ def get_vector_store():
     if not os.path.exists(STORAGE_DIR):
         vector_store = SimpleVectorStore()
```
marcusschiesser, line 9:
```suggestion
        # Vector store hasn't been persisted before, create a new one
        vector_store = SimpleVectorStore()
```

templates/components/vectordbs/python/none/vectordb.py
```diff
@@ -8,6 +8,8 @@ def get_vector_store():
...
     else:
-        vector_store = SimpleVectorStore.from_persist_dir(STORAGE_DIR)
+        vector_store = SimpleVectorStore.from_persist_dir(
```
marcusschiesser, line 11:
```suggestion
        # Vector store has already been persisted before at STORAGE_DIR - load it
        vector_store = SimpleVectorStore.from_persist_dir(
```

templates/types/streaming/fastapi/app/engine/generate.py
```diff
@@ -54,14 +57,24 @@ def generate_datasource():
...
+                docstore=docstore,
+                vector_store=vector_store,
+            ),
+            store_nodes_override=True,
+        )
+        # SimpleDocumentStore keeps the data in memory only - needs to be persisted explicitly
```
marcusschiesser, line 73: @leehuwuj you made a mistake appying this change from me

templates/types/streaming/fastapi/app/engine/generate.py
```diff
@@ -54,14 +57,24 @@ def generate_datasource():
     ingestion_pipeline.vector_store = vector_store
 
     # Run the ingestion pipeline and store the results
-    ingestion_pipeline.run(show_progress=True, documents=documents)
+    nodes = ingestion_pipeline.run(show_progress=True, documents=documents)
 
-    # Default vector store only keeps data in memory, so we need to persist it
-    # Can remove if using a different vector store
+    # SimpleVectorStore keeps the data in memory only - needs to be persisted explicitly
```
marcusschiesser, line 62: i think we should put this whole code in its own `persist` function

templates/types/streaming/fastapi/app/engine/generate.py
leehuwuj: Ah, i thought you said applying `storage_context.persist` for other vector stores below. But this condition is already persited by `index.storagecontext.persist(STORAGE_DIR)`. which is the same as your idea right?

templates/types/streaming/fastapi/app/engine/generate.py
leehuwuj: great catch! i just applied the suggestion automatically to the wrong line, let me update it again

templates/types/streaming/fastapi/app/engine/generate.py
leehuwuj: Yes, i also think that we should move the persist to another function.

templates/components/vectordbs/python/none/vectordb.py
leehuwuj: Just removed, we don't need stores_text = true

leehuwuj:
Closed due to the ingestion pipeline issue. Move embedding dimension fixing into other PR.

PR #74, opened 2024-05-03
Use ingestion pipeline

leehuwuj:
This PR is not ready to merge because we're having issue with llama_index ingestion pipeline.
Experiment notebook: https://drive.google.com/file/d/1b19t5n6jtfex2UuKCRgxzNaCzk8a1ady/view?usp=sharing

## Summary by CodeRabbit

- **New Features**
  - Introduced logging functionality for enhanced monitoring.
  - Added functions for generating and persisting data indexes.
  - Implemented dynamic storage directory setting via environment variables.

- **Improvements**
  - Enhanced exception handling for empty data directories.
  - Improved data ingestion pipeline and storage persistence mechanisms.

- **Refactor**
  - Updated storage directory logic to use environment variables.
  - Reorganized functions for better modularity and clarity.


templates/types/streaming/fastapi/app/engine/constants.py
```diff
@@ -0,0 +1 @@
+STORAGE_DIR = "storage"
```
leehuwuj, line 1: we've only got storage_dir as a constant right now. Switching it to environment variable is not a bad idea

templates/types/streaming/fastapi/app/engine/generate.py
```diff
@@ -0,0 +1,79 @@
...
+logger = logging.getLogger()
+
+
+def get_doc_store():
+    if not os.path.exists(STORAGE_DIR):
+        docstore = SimpleDocumentStore()
+        return docstore
+    else:
+        logger.info(f"Loading existing docstore from {STORAGE_DIR}")
+        return SimpleDocumentStore.from_persist_dir(STORAGE_DIR)
```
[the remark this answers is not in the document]
marcusschiesser: @leehuwuj that looks like a good idea

templates/components/loaders/python/file.py
```diff
@@ -27,11 +30,26 @@ def llama_parse_parser():
...
+
+        _, _, exc_traceback = sys.exc_info()
+        function_name = traceback.extract_tb(exc_traceback)[-1].name
+        if function_name == "_add_files":
+            logger.warning(
+                f"Failed to load file documents, error message: {e} . Return as empty document list."
+            )
+            return []
+        else:
+            raise e
```
marcusschiesser, line 55: can't we just log and raise?

templates/types/streaming/fastapi/app/engine/generate.py
```diff
@@ -0,0 +1,79 @@
...
```
marcusschiesser, line 29: can we just use this line? (as we won't use an in memory docstore)

templates/types/streaming/fastapi/app/engine/index.py
```diff
@@ -0,0 +1,16 @@
...
+from app.engine.vectordb import get_vector_store
+
+
+logger = logging.getLogger("uvicorn")
+
+
+def get_index():
+    logger.info("Connecting vector store...")
+    store = get_vector_store()
+    index = VectorStoreIndex.from_vector_store(store)
```
marcusschiesser, line 14: is this using the doc_store?

templates/types/streaming/fastapi/app/engine/generate.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def generate_datasource():
+    init_settings()
+    logger.info("Generate index for the provided data")
+
+    # Get the stores and documents or create new ones
+    documents = get_documents()
+    docstore = get_doc_store()
+    vector_store = get_vector_store()
```
marcusschiesser, line 67: where is the code for  getting get_vector_store in the none case

templates/types/streaming/fastapi/app/engine/constants.py
marcusschiesser: yes or inlining constant if it's just used once

templates/types/streaming/fastapi/app/engine/generate.py
leehuwuj: I'm letting the none case use its own generate script as is (not using the ingestion pipeline). So, the files in vectordbs/none folder will override this.

templates/types/streaming/fastapi/app/engine/index.py
leehuwuj: No, we are only using the doc_store for deduplication in generating the index.

marcusschiesser: yes but the doc_store should also be used if the index is read

templates/types/streaming/fastapi/app/engine/generate.py
marcusschiesser: ah got it! good idea but a bit brittle - do we have vector/none/python in e2e tests?

templates/types/streaming/fastapi/app/engine/generate.py
leehuwuj: afaik, the ingestion pipeline won't initialize an empty docstore (it will switch to deduplicate strategy if docstore is not provided), and we cannot initialize an empty docstore from an empty storage path.

templates/components/loaders/python/file.py
leehuwuj: Actually, i added this code to cover a case where the data folder is empty, in that case, we should return an empty list.

templates/types/streaming/fastapi/app/engine/generate.py
leehuwuj: Yes, we are using none vectordb with example file option.

templates/components/loaders/python/file.py
marcusschiesser: makes sense but then a comment would be great :)

PR #94, opened 2024-05-20
feat: implement interpreter tool

thucpn:
image
image

## Summary by CodeRabbit

- **New Features**
  - Introduced the "Interpreter" tool for executing Python code in a secure environment, supporting various output formats such as HTML, markdown, and images.
  - Added new routes and endpoints for serving static files in Express and FastAPI applications.
  - Enhanced API functionality to retrieve file data from different folders based on URL paths.

- **Bug Fixes**
  - Improved error handling in the `createChatEngine` function to log and rethrow errors when loading tools.

- **Dependencies**
  - Added `@e2b/code-interpreter` as a new dependency in the `package.json` for Express and Next.js templates.

- **Chores**
  - Updated GitHub Actions workflows to use the latest versions of `actions/setup-python` and `actions/upload-artifact`.
  - Added `tool-output/` directory to `.gitignore` in the FastAPI template.


helpers/tools.ts
```diff
@@ -62,6 +62,13 @@ export const supportedTools: Tool[] = [
     supportedFrameworks: ["fastapi", "express", "nextjs"],
     type: ToolType.LOCAL,
   },
+  {
+    display: "Interpreter",
+    name: "interpreter",
+    dependencies: [],
+    supportedFrameworks: ["fastapi", "express", "nextjs"],
+    type: ToolType.LOCAL,
+  },
```
leehuwuj, line 71: Do we support ToolSpec in TS version? 
E2B interpreter supports executing python and javascript and it requires API key so i think it's better to implement a tool spec for it.
```suggestion
  {
    display: "Code Interpreter",
    name: "e2b.code_interpreter",
    dependencies: [],
    supportedFrameworks: ["fastapi", "express", "nextjs"],
    type: ToolType.LOCAL,
  },
```

templates/components/engines/typescript/agent/tools/interpreter.ts
```diff
@@ -0,0 +1,171 @@
...
+
+const DEFAULT_META_DATA: ToolMetadata<JSONSchemaType<InterpreterParameter>> = {
+  name: "interpreter",
+  description: `
+  - You are a Python interpreter.
+  - You are given tasks to complete and you run python code to solve them.
+  - The python code runs in a Jupyter notebook. Every time you call \`interpreter\` tool, the python code is executed in a separate cell. It's okay to make multiple calls to \`interpreter\`.
+  - Display visualizations using matplotlib or any other visualization library directly in the notebook. Shouldn't save the visualizations to a file, just return the base64 encoded data.
+  - You can install any pip package (if it exists) if you need to but the usual packages for data analysis are already preinstalled.
+  - You can run any python code you want in a secure environment.
```
leehuwuj, line 48: This seems more like a system prompt than a function description.

marcusschiesser: i agree, this is the tool description that is used by the LLM to call the tool, should be something simple like:
`Execute python code in a Jupyter notebook cell and returns any result, stdout, stderr, display_data, and error.`
The rest should be in the system prompt of the agent: I guess this means that we need a different system prompt in the .env file if we select the code interpreter tool

for the difference between the system prompt and the tool description, see prompting at https://e2b.dev/docs/hello-world/js

helpers/tools.ts
marcusschiesser: yes, usually the tool should be configured with E2B's API key. But usually that seems to be in the E2B_API_KEY variable. So for the dev it would be better to edit `E2B_API_KEY` in the env file.
Can we support env variable generation if a specific tool is selected?

templates/components/engines/typescript/agent/tools/interpreter.ts
```diff
@@ -0,0 +1,171 @@
...
+        type: "string",
+        description: "The python code to execute in a single cell.",
+      },
+    },
+    required: ["code"],
+  },
+};
+
+export class InterpreterTool implements BaseTool<InterpreterParameter> {
+  private readonly outputDir = "data";
```
marcusschiesser, line 63: we can't use data dir for the output of the interpreter 
(because running generate again would index that stuff)
any ideas? 
how about using a sub directory or a different directory or we just use base64 encoded strings without filesystem?

templates/components/engines/typescript/agent/tools/interpreter.ts
```diff
@@ -0,0 +1,171 @@
...
+  private readonly dataAPI = "/api/data/";
+  private apiKey?: string;
+  metadata: ToolMetadata<JSONSchemaType<InterpreterParameter>>;
+  codeInterpreter?: CodeInterpreter;
+
+  constructor(params?: InterpreterToolParams) {
```
marcusschiesser, line 69: if an agent is using this tool, how often is the tool initiated? is it the lifetime of the task or of the agent?

templates/types/streaming/nextjs/package.json
```diff
@@ -34,7 +34,8 @@
     "supports-color": "^8.1.1",
     "tailwind-merge": "^2.1.0",
     "vaul": "^0.9.1",
-    "@llamaindex/pdf-viewer": "^1.1.1"
+    "@llamaindex/pdf-viewer": "^1.1.1",
+    "@e2b/code-interpreter": "^0.0.5"
```
marcusschiesser, line 38: what about express?

templates/types/streaming/nextjs/app/components/ui/chat/markdown.tsx
```diff
@@ -28,8 +29,13 @@ const preprocessLaTeX = (content: string) => {
   return inlineProcessedContent;
 };
 
+const preprocessUrl = (content: string) => {
+  return replaceAttachmentUrl(content);
```
marcusschiesser, line 33: this is showing the inline images in a message see?

image

templates/types/streaming/nextjs/app/components/ui/chat/widgets/InterpreterCard.tsx
```diff
@@ -0,0 +1,34 @@
...
+    stdout: string[];
+  };
+}
+
+export function InterpreterCard({ data }: { data: InterpreterData }) {
+  const { isError, extraResult } = data;
+  if (isError || !extraResult.length) return null;
+  return (
+    <div className="space-x-2">
+      <span className="font-semibold">Output Files:</span>
```
marcusschiesser, line 21: image
our tool output widget is just linking the output files?

templates/components/engines/typescript/agent/tools/interpreter.ts
marcusschiesser: I think we don't need the tool system prompt during run-time, just at code generation time.
So how about: 
If the user selects a specific tool, we suggest to use a specific system prompt.

Means you can add the tool system prompt to helpers/tools.ts
and when you generate the env variable you use the system prompt of the selected tool (for now you can just concat if the user selects more than one tool with a system prompt)

templates/components/engines/typescript/agent/tools/interpreter.ts
marcusschiesser: ok, but then we should use another folder

templates/components/engines/typescript/agent/tools/interpreter.ts
marcusschiesser: i think the problem is that the agent class is stateful. what happens if we reuse the agent instance between requests?
ideally, each user probably should have its own agent? let's deal with this issue in another PR and just test if it works with a codeinterpreter as is

templates/types/streaming/nextjs/app/components/ui/chat/markdown.tsx
marcusschiesser: yes, but before and after the widget there might be some markdown text from the LLM

how about:
1. if we see an attachment in the markdown, we replace it with rendering an image widget
2. this image widget shows the image AND allows us to do something with the file (e.g. download it - the functionality of the current interpreter widget)

but for the start (this PR) we can just show the image inline and not use the interpreter widget at all (keep the example simple!)

templates/types/streaming/nextjs/app/components/ui/chat/widgets/InterpreterCard.tsx
marcusschiesser: yes probably a next step (see above), I'll make a note for a "inline widget ticket"

templates/types/streaming/nextjs/app/components/ui/chat/chat-tools.tsx
```diff
@@ -20,6 +20,10 @@ export default function ChatTools({ data }: { data: ToolData }) {
     case "get_weather_information":
       const weatherData = toolOutput.output as unknown as WeatherData;
       return <WeatherCard data={weatherData} />;
+    // We will support displaying interpreter output later
+    // case "interpreter":
+    //   const interpreterData = toolOutput.output as unknown as InterpreterData;
+    //   return <InterpreterCard data={interpreterData} />;
```
marcusschiesser, line 26: remove from this PR

templates/components/engines/typescript/agent/tools/interpreter.ts
```diff
@@ -0,0 +1,160 @@
...
+
+export type InterpreterExtraResult = {
+  type: InterpreterExtraType;
+  filename: string;
+};
+
+const DEFAULT_META_DATA: ToolMetadata<JSONSchemaType<InterpreterParameter>> = {
+  name: "interpreter",
+  description:
+    "Execute python code in a Jupyter notebook cell and returns any result, stdout, stderr, display_data, and error.",
```
marcusschiesser, line 42:
```suggestion
    "Execute python code in a Jupyter notebook cell and return any result, stdout, stderr, display_data, and error.",
```

templates/types/streaming/nextjs/app/components/ui/lib/url.ts
```diff
@@ -9,3 +9,16 @@ export const getStaticFileDataUrl = (filename: string) => {
...
 };
+
+// replace all attachment:// url with /api/data/
+export const replaceAttachmentUrl = (content: string) => {
+  const isUsingBackend = !!process.env.NEXT_PUBLIC_CHAT_API;
+  if (isUsingBackend) {
+    const backendOrigin = new URL(process.env.NEXT_PUBLIC_CHAT_API!).origin;
+    return content.replace(
+      /attachment:\/\//g,
+      `${backendOrigin}/api/${STORAGE_FOLDER}/`,
```
marcusschiesser, line 20: how is this working if images are saved in `tool-output`?

helpers/tools.ts
```diff
@@ -99,6 +99,11 @@ export const supportedTools: Tool[] = [
         description:
           "E2B_API_KEY key is required to run code interpreter tool. Get it here: https://e2b.dev/docs/getting-started/api-key",
       },
+      {
+        name: "E2B_FILESYSTEM_ORIGIN",
+        description:
+          "E2B_FILESYSTEM_ORIGIN is the origin of the filesystem server. It's ussually the same as your backend origin. Example: http://localhost:3000",
```
marcusschiesser, line 105:
```suggestion
          "E2B_FILESYSTEM_ORIGIN is the origin of the filesystem server. It's usually the same as your backend origin. Example: http://localhost:3000",
```

templates/types/streaming/nextjs/app/api/data/[path]/route.ts
```diff
@@ -21,7 +21,11 @@ export async function GET(
   }
 
   try {
-    const filePath = path.join(process.cwd(), "data", slug);
+    let retrievedFolder = "data"; // default folder
```
Comment (marcusschiesser) [inline templates/types/streaming/nextjs/app/api/data/[path]/route.ts:24]: can we set the route to /api/[path]/route.ts and filter out requests that are not data nor tool-output?

Comment (marcusschiesser) [reply templates/types/streaming/nextjs/app/api/data/[path]/route.ts:24]: or maybe we can use the /api/files/[path]/route.ts endpoint instead (and [path] can contain data or tool-output? so we can easily extend it for other files?

templates/components/engines/typescript/agent/tools/interpreter.ts
```diff
@@ -0,0 +1,174 @@
...
+  private getOutputPath(filename: string): string {
+    // if outputDir doesn't exist, create it
+    if (!fs.existsSync(this.outputDir)) {
+      fs.mkdirSync(this.outputDir, { recursive: true });
+    }
+    return path.join(this.outputDir, filename);
+  }
+
+  private getFileUrl(filename: string): string {
+    return `${this.fileSystemOrigin}/api/tool-output/${filename}`;
```
marcusschiesser, line 172:
```suggestion
    return `${this.fileSystemOrigin}/api/${this.outputDir}/${filename}`;
```

marcusschiesser: maybe we can then also store /api/ in `this.fileSystemOrigin`?
we should better call it `E2B_FILESERVER_URL_PREFIX` then I guess

templates/types/streaming/fastapi/main.py
```diff
@@ -38,10 +38,8 @@ async def redirect_to_docs():
         return RedirectResponse(url="/docs")
 
 
-if os.path.exists("data"):
-    app.mount("/api/data", StaticFiles(directory="data"), name="static")
-
-app.mount("/api/tool-output", StaticFiles(directory="tool-output"), name="static")
+app.mount("/api/files/data", StaticFiles(directory="data"), name="data-static")
+app.mount("/api/files/tool-output", StaticFiles(directory="tool-output"), name="tool-output-static")
```
marcusschiesser, line 42: don't remove the path check and don't add tool output yet (no python support in this PR)
```suggestion
if os.path.exists("data"):
    app.mount("/api/files/data", StaticFiles(directory="data"), name="data-static")
```

templates/types/streaming/nextjs/next.config.mjs
```diff
@@ -5,4 +5,14 @@ import webpack from "./webpack.config.mjs";
 const nextConfig = JSON.parse(fs.readFileSync("./next.config.json", "utf-8"));
 nextConfig.webpack = webpack;
 
+// redirect /api/tool-output/ to api/data/ with same slug
+nextConfig.rewrites = async () => {
```
marcusschiesser, line 9: needed?

templates/types/streaming/fastapi/gitignore
```diff
@@ -1,3 +1,5 @@
 __pycache__
 storage
 .env
+
+tool-output/
```
marcusschiesser, line 5: not needed yet

templates/types/streaming/express/index.ts
```diff
@@ -31,8 +31,8 @@ if (isDevelopment) {
   console.warn("Production CORS origin not set, defaulting to no CORS.");
 }
 
-app.use("/api/data", express.static("data"));
-app.use("/api/tool-output", express.static("tool-output"));
+app.use("/api/files/data", express.static("data"));
```
marcusschiesser, line 34: where are the changes in the source node viewer that we changed /api/data to api/files/data?

templates/components/engines/typescript/agent/tools/interpreter.ts
```diff
@@ -169,6 +169,6 @@ export class InterpreterTool implements BaseTool<InterpreterParameter> {
   }
 
   private getFileUrl(filename: string): string {
-    return `${this.fileSystemOrigin}/api/tool-output/${filename}`;
+    return `${this.fileSystemOrigin}/api/${this.outputDir}/${filename}`;
```
marcusschiesser, line 172:
```suggestion
    return `${this.fileServerURLPrefix}/${this.outputDir}/${filename}`;
```

helpers/tools.ts
```diff
@@ -100,9 +100,9 @@ export const supportedTools: Tool[] = [
           "E2B_API_KEY key is required to run code interpreter tool. Get it here: https://e2b.dev/docs/getting-started/api-key",
       },
       {
-        name: "E2B_FILESYSTEM_ORIGIN",
+        name: "E2B_FILESERVER_URL_PREFIX",
         description:
-          "E2B_FILESYSTEM_ORIGIN is the origin of the filesystem server. It's ussually the same as your backend origin. Example: http://localhost:3000",
+          "E2B_FILESERVER_URL_PREFIX is the origin of the filesystem server. It's usually the same as your backend origin. Example: http://localhost:3000",
```
marcusschiesser, line 105: we should pre-generate this as it's complicated:
if backend:
http://localhost:8000/api/files
else:
http://localhost:3000/api/files

```suggestion
          "E2B_FILESERVER_URL_PREFIX is the URL prefix of the server storing the images generated by the interpreter.",
```

.changeset/gentle-knives-unite.md
```diff
@@ -0,0 +1,5 @@
+---
+"create-llama": patch
+---
+
+feat: implement interpreter tool
```
marcusschiesser, line 5:
```suggestion
Add interpreter tool for TS using e2b.dev
```

helpers/tools.ts
marcusschiesser: no worries, i'll add it here

templates/types/streaming/nextjs/app/components/ui/lib/url.ts
```diff
@@ -1,11 +1,11 @@
...
 
-export const getStaticFileDataUrl = (filename: string) => {
+export const getStaticFileDataUrl = (filePath: string) => {
   const isUsingBackend = !!process.env.NEXT_PUBLIC_CHAT_API;
-  const fileUrl = `/api/${STORAGE_FOLDER}/${filename}`;
+  const fileUrl = `${staticFileAPI}/${filePath}`;
   if (isUsingBackend) {
     const backendOrigin = new URL(process.env.NEXT_PUBLIC_CHAT_API!).origin;
-    return `${backendOrigin}/${fileUrl}`;
+    return `${backendOrigin}${fileUrl}`;
```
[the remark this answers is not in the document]
marcusschiesser: ah, it's not filePath sorry

Issue #43, opened 2024-04-09
Supabase as Vector DB FATAL:  Max client connections reached


meera:
Project generated with create-llama. 
Works most of the time - after a while  throws up this error. 

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/Me/Library/Caches/pypoetry/virtualenvs/app-BCQ_KSqF-py3.11/lib/python3.11/site-packages/uvicorn/protocols/http/httptools_impl.py", line 426, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/Me/Library/Caches/pypoetry/virtualenvs/app-BCQ_KSqF-py3.11/lib/python3.11/site-packages/uvicorn/middleware/proxy_headers.py", line 84, in __call__
    return await self.app(scope, receive, send)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/Me/Library/Caches/pypoetry/virtualenvs/app-BCQ_KSqF-py3.11/lib/python3.11/site-packages/fastapi/applications.py", line 1054, in __call__
    await super().__call__(scope, receive, send)
  File "/Users/Me/Library/Caches/pypoetry/virtualenvs/app-BCQ_KSqF-py3.11/lib/python3.11/site-packages/starlette/applications.py", line 123, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/Users/Me/Library/Caches/pypoetry/virtualenvs/app-BCQ_KSqF-py3.11/lib/python3.11/site-packages/starlette/middleware/errors.py", line 186, in __call__
    raise exc
  File "/Users/Me/Library/Caches/pypoetry/virtualenvs/app-BCQ_KSqF-py3.11/lib/python3.11/site-packages/starlette/middleware/errors.py", line 164, in __call__
    await self.app(scope, receive, _send)
  File "/Users/Me/Library/Caches/pypoetry/virtualenvs/app-BCQ_KSqF-py3.11/lib/python3.11/site-packages/starlette/middleware/cors.py", line 83, in __call__
    await self.app(scope, receive, send)
  File "/Users/Me/Library/Caches/pypoetry/virtualenvs/app-BCQ_KSqF-py3.11/lib/python3.11/site-packages/starlette/middleware/exceptions.py", line 62, in __call__
[...]

PR #95, opened 2024-05-21
fix: Reuse vector store to avoid recreating sqlalchemy engine

leehuwuj:
Fix: https://github.com/run-llama/create-llama/issues/43

## Summary by CodeRabbit

- **New Features**
  - Introduced caching mechanism for `get_storage_context` to enhance performance.
  
- **Performance Improvements**
  - Added a global variable to cache `PGVectorStore` instance, reducing redundant object creation.
  - Implemented caching for `get_storage_context` using `cachetools` to improve efficiency.

- **Dependencies**
  - Added `cachetools` version `^5.3.3` to project dependencies.

### DIFF
--- a/templates/components/vectordbs/python/none/index.py
+++ b/templates/components/vectordbs/python/none/index.py
@@ -1,10 +1,20 @@
-import logging
 import os
+import logging
+from datetime import timedelta
 
+from cachetools import cached, TTLCache
 from llama_index.core.storage import StorageContext
 from llama_index.core.indices import load_index_from_storage
 
 logger = logging.getLogger("uvicorn")
+
+
+@cached(
+    TTLCache(maxsize=10, ttl=timedelta(minutes=5).total_seconds()),
+    key=lambda *args, **kwargs: "global_storage_context",
+)
+def get_storage_context(persist_dir: str) -> StorageContext:
+    return StorageContext.from_defaults(persist_dir=persist_dir)
 
 
 def get_index():
@@ -14,7 +24,7 @@ def get_index():
         return None
     # load the existing index
     logger.info(f"Loading index from {storage_dir}...")
-    storage_context = StorageContext.from_defaults(persist_dir=storage_dir)
+    storage_context = get_storage_context(storage_dir)
     index = load_index_from_storage(storage_context)
     logger.info(f"Finished loading index from {storage_dir}")
     return index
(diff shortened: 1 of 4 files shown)


templates/components/vectordbs/python/pg/vectordb.py
```diff
@@ -5,26 +5,33 @@
 PGVECTOR_SCHEMA = "public"
 PGVECTOR_TABLE = "llamaindex_embedding"
 
+vector_store: PGVectorStore = None
+
 
 def get_vector_store():
```
marcusschiesser, line 11: sec insights is using a TTL cache for storage context (see https://github.com/run-llama/sec-insights/blob/94ec3507ca5ea8f67c52d82ef69d18953b2766ad/backend/app/chat/engine.py#L121)

1. can we also use that for vector_store? sounds better than having a global vector_store
2. if we don't create a vector store per request, the vector_store must be thread safe. is it?
3. does it make sense to use this ttl cache also for the whole storage context?

leehuwuj: 1. Yes, we can use the cache, just simple cache is enough. 
2. It must be thread safe but I think in fastapi app each request is already handling by a uvicorn worker, it's working now so it should be working with the update.
3. We’re not using all of the storage context store, but it’s worth applying it to other vector stores too.
How about using a singleton object since it's easier to handle (lazy init,..).

templates/components/vectordbs/python/none/index.py
```diff
@@ -1,20 +1,30 @@
-import logging
 import os
+import logging
+from datetime import timedelta
 
+from cachetools import cached, TTLCache
```
marcusschiesser, line 5: not a new dependency?

templates/components/vectordbs/python/pg/vectordb.py
```diff
@@ -5,26 +5,33 @@
...
+        # Update the configured scheme with the psycopg2 and asyncpg schemes
+        original_scheme = urlparse(original_conn_string).scheme + "://"
+        conn_string = original_conn_string.replace(
+            original_scheme, "postgresql+psycopg2://"
+        )
+        async_conn_string = original_conn_string.replace(
+            original_scheme, "postgresql+asyncpg://"
+        )
+
+        vector_store = PGVectorStore(
```
marcusschiesser, line 29: what happens if pg disconnects? how do we reconnect?

templates/components/vectordbs/python/none/index.py
leehuwuj: Ah great catch!

templates/components/vectordbs/python/pg/vectordb.py
leehuwuj: The connection session and connection pool are managed by the sqlalchemy so it'll reconnect automatically

marcusschiesser (approved):

PR #100, opened 2024-05-27
feat: add support for FastAPI in code interpreter tool

leehuwuj:


templates/components/engines/python/agent/tools/interpreter.py
```diff
@@ -0,0 +1,134 @@
...
+                    output = E2BToolOutput(is_error=False, logs=exec.logs, results=[])
+                else:
+                    results = self.parse_result(exec.results[0])
+                    output = E2BToolOutput(
+                        is_error=False, logs=exec.logs, results=results
+                    )
+            return output.dict()
+
+
+def code_interpret(code: str) -> Dict:
```
leehuwuj, line 112: The api_key and fileserver_url_prefix are loaded from env so I created a dedicated function to initialize the E2BCodeInterpreter and we will use the function as a function tool.

marcusschiesser: good idea

helpers/index.ts
```diff
@@ -171,6 +171,11 @@ export const installTemplate = async (
         );
       }
     }
+
+    // Create tool-output directory
+    if (props.tools && props.tools.length > 0) {
```
marcusschiesser, line 176: i see that python tool is generating the output dir if it doesn't exist - how about we add this to TS too?

leehuwuj: I think this change is already applied to TS.

PR #104, opened 2024-05-29
feat: display conversation starter from backend env

thucpn:
## Summary by CodeRabbit

- **New Features**
  - Introduced conversation starter questions in the chat interface.
  - Added a new API endpoint in the Next.js application to fetch chat configuration data.
  - Implemented a custom hook `useClientConfig` to retrieve chat configuration.

- **Enhancements**
  - Updated chat components to display starter questions as buttons when no messages are present.
  - Added the ability to append messages in the chat interface.

- **Bug Fixes**
  - Ensured environment variables specific to the "nextjs" framework are correctly loaded.

- **Documentation**
  - Added a changeset file describing the new functionality for conversation starters.


helpers/env-variables.ts
```diff
@@ -185,6 +185,12 @@ const getModelEnvs = (modelConfig: ModelConfig): EnvVar[] => {
       description: "Dimension of the embedding model to use.",
       value: modelConfig.dimensions.toString(),
     },
+    {
+      name: "CONVERSATION_STARTERS",
+      description: "The questions to help users get started (multi-line).",
+      value:
+        '"\nWhat is the weather today?\nWhat is the capital of France?\nWhat is the largest mammal in the world?\nWhat is the largest planet in the solar system?\n"',
```
marcusschiesser, line 192: I would default to not showing any options

templates/types/streaming/nextjs/app/components/ui/chat/use-config.ts
```diff
@@ -0,0 +1,27 @@
+"use client";
+
+import { useEffect, useMemo, useState } from "react";
+
+export interface ChatConfig {
```
marcusschiesser, line 5: How about adding the vale of NEXT_PUBLIC_CHAT_API to this config object?

templates/types/streaming/fastapi/app/api/routers/chat.py
```diff
@@ -168,3 +171,14 @@ async def chat_request(
         result=_Message(role=MessageRole.ASSISTANT, content=response.response),
         nodes=_SourceNodes.from_source_nodes(response.source_nodes),
     )
+
+
+@r.get("/config")
```
leehuwuj, line 176: We have a Swagger document so i think it's best to create a config pydantic model so that the user can understand what the output is (check the chat and message model above)

templates/types/streaming/nextjs/app/components/ui/chat/use-config.ts
leehuwuj: Because this the model for the data from BE so i think the BE doesn't need to know what is the NEXT_PUBLIC_CHAT_API

templates/types/streaming/fastapi/app/api/routers/chat.py
```diff
@@ -168,3 +175,14 @@ async def chat_request(
...
+async def chat_config() -> _ChatConfig:
+    conversation_starters = os.getenv("CONVERSATION_STARTERS")
+    starter_questions = (
+        conversation_starters.strip().split("\n") if conversation_starters else None
```
leehuwuj, line 184: @thucpn how about we do not convert the starter_questions if the `CONVERSATION_STARTERS` is an empty string? It's normal that the user set it as empty string instead of letting it as null

templates/types/streaming/fastapi/app/api/routers/chat.py
```diff
@@ -74,6 +77,10 @@ class _Result(BaseModel):
     nodes: List[_SourceNodes]
 
 
+class _ChatConfig(BaseModel):
+    starterQuestions: Optional[List[str]]
+
```
leehuwuj, line 82: In python, we use snake case instead of camel case. You can try advance feature from pydantic to alias the field name and add the field description too.
```suggestion
class _ChatConfig(BaseModel):
    starter_questions: Optional[List[str]] = Field(
        default=None,
        description="List of starter questions",
    )

    class Config:
        json_schema_extra = {
            "example": {
                "starterQuestions": [
                    "What standards for letters exist?",
                    "What are the requirements for a letter to be considered a letter?",
                ]
            }
        }
        alias_generator = to_camel
```
and don't forget to update the import:
```python
from pydantic import BaseModel, Field
from pydantic.alias_generators import to_camel
```

templates/types/streaming/nextjs/app/components/ui/chat/use-config.ts
marcusschiesser: This is client code right? I mean we should centralize the usage of NEXT_PUBLIC_CHAT_API at one point

marcusschiesser (approved):

PR #113, opened 2024-06-05
Support uploading file to e2b code interpreter tool

leehuwuj:


templates/components/engines/python/agent/tools/interpreter.py
```diff
@@ -14,14 +14,15 @@
 
 class InterpreterExtraResult(BaseModel):
     type: str
-    filename: str
-    url: str
+    content: Optional[str] = None
```
leehuwuj, line 17: The e2b result is not only a file but also a raw text, html markdown text.

templates/components/engines/python/agent/tools/interpreter.py
```diff
@@ -108,9 +125,15 @@ def interpret(self, code: str) -> E2BToolOutput:
             return output
 
 
-def code_interpret(code: str) -> Dict:
+def code_interpret(code: str, local_file_path: str) -> Dict:
```
leehuwuj, line 128: - code: the code to be executed.
- local_file_path: the uploaded file location for transfer to the e2b sandbox.

templates/components/engines/python/agent/tools/interpreter.py
```diff
@@ -108,9 +125,15 @@ def interpret(self, code: str) -> E2BToolOutput:
...
     """
-    Execute python code in a Jupyter notebook cell and return any result, stdout, stderr, display_data, and error.
+    Use this tool to analyze the provided data in a sandbox environment.
+    The tool will:
+        1. Upload the provided file from local to the sandbox. The uploaded file path will be /home/user/{filename}
+        2. Execute python code in a Jupyter notebook cell to analyze the uploaded file in the sandbox.
+        3. Get the result from the code in stdout, stderr, display_data, and error.
+    You must to provide the code and the provided file path to run this tool.
+    Your code should read the file from the sandbox path /home/user/{filename}.
```
leehuwuj, line 136: An additional note for AI to use the path in the sandbox environment

templates/components/engines/python/agent/tools/interpreter.py
```diff
@@ -72,31 +74,46 @@ def parse_result(self, result) -> List[InterpreterExtraResult]:
...
+                        output.append(InterpreterExtraResult(type=ext, content=data))
         except Exception as error:
             logger.error("Error when saving data to disk", error)
 
         return output
 
-    def interpret(self, code: str) -> E2BToolOutput:
+    def interpret(self, code: str, file_path: Optional[str] = None) -> E2BToolOutput:
         with CodeInterpreter(api_key=self.api_key) as interpreter:
+            # Upload file to E2B sandbox
```
leehuwuj, line 101: Always upload the file for each code interpreter calling. We can improve this by introducing a chat session feature, then each session can use its own sandbox environment.

marcusschiesser: makes sense. I think this is a general concept we have to add later, e.g. we also need to store the agent's memory in that session store 
as we generate demo code - for now we just have to add an explanation to the user

templates/components/engines/python/agent/tools/interpreter.py
marcusschiesser: what is for the case that the user doesn't upload a file?

leehuwuj: I also thought about that, we can use local_file_path as optional argument for the case the code does not use any data, but that doesn't seem practical.

templates/components/engines/python/agent/tools/interpreter.py
marcusschiesser: @leehuwuj is that not independant of using files?

templates/components/engines/typescript/agent/tools/interpreter.ts
```diff
@@ -119,18 +142,26 @@ export class InterpreterTool implements BaseTool<InterpreterParameter> {
     try {
       const formats = res.formats(); // formats available for the result. Eg: ['png', ...]
       const base64DataArr = formats.map((f) => res[f as keyof Result]); // get base64 data for each format
+      console.log("data", base64DataArr);
```
marcusschiesser, line 145: no output

templates/components/engines/python/agent/tools/interpreter.py
marcusschiesser: what do you mean by "not practical"?

templates/components/engines/typescript/agent/tools/interpreter.ts
```diff
@@ -88,11 +100,22 @@ export class InterpreterTool implements BaseTool<InterpreterParameter> {
     return this.codeInterpreter;
   }
 
-  public async codeInterpret(code: string): Promise<InterpreterToolOutput> {
+  public async codeInterpret(
+    code: string,
+    localFilePath: string,
```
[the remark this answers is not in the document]
marcusschiesser: yes i also had that question for python, see https://github.com/run-llama/create-llama/pull/113/files#r1627342305

leehuwuj:
Closed in favor of: https://github.com/run-llama/create-llama/pull/355

PR #114, opened 2024-06-05
feat: display attachments in user messages

thucpn:
## Summary by CodeRabbit

- **New Features**
  - Added a new optional function `setInput` to the chat input component for better input management.
  - Introduced an `isHidden` property to conditionally hide chat message components based on user roles.

- **Refactor**
  - Updated CSV content rendering layout for improved display.
  - Refactored the `UploadCsvPreview` component to use a `Drawer` for better organization and interaction.

- **Chores**
  - Dynamic setting of `maxTokens` in the OpenAI initialization based on environment variables for better configurability.


templates/types/streaming/nextjs/app/components/ui/chat/chat-message.tsx
```diff
@@ -60,8 +61,9 @@ function ChatMessageContent({
 
   const contents: ContentDisplayConfig[] = [
     {
-      order: -4,
+      order: 1,
       component: imageData[0] ? <ChatImage data={imageData[0]} /> : null,
+      isHidden: message.role === "assistant",
```
marcusschiesser, line 66: isn't that a backend problem that the backend is still sending this annotation for assitant's messages?

templates/types/streaming/nextjs/app/components/ui/chat/widgets/CsvDialog.tsx
```diff
@@ -22,22 +23,26 @@ export default function CsvDialog(props: CsvDialogProps) {
...
-            <Image
-              className="h-full w-auto"
-              priority
-              src={SheetIcon}
-              alt="SheetIcon"
-            />
+        {props.trigger ? (
+          <div className="cursor-pointer">{props.trigger}</div>
+        ) : (
+          <div
```
marcusschiesser, line 29: This is basically another `UploadCsvPreview` - one without a trigger - can we merge this code? I think anyhow that the two Files should be rendered similar (inside chat input and the user's message)

templates/types/streaming/nextjs/app/components/ui/chat/chat-input.tsx
```diff
@@ -19,30 +22,70 @@ export default function ChatInput(
...
   );
 
-  const onSubmit = (e: React.FormEvent<HTMLFormElement>) => {
+  const getAttachments = () => {
+    if (!imageUrl && files.length === 0) return undefined;
+    const annotations: MessageAnnotation[] = [];
     if (imageUrl) {
-      props.handleSubmit(e, {
-        data: { imageUrl: imageUrl },
+      annotations.push({
```
marcusschiesser, line 38: AFAIK the server can also send user annotations. Now we're mixing creating annotations on the server side and on the client side, which might be confusing

templates/types/streaming/nextjs/app/components/ui/chat/chat-message.tsx
marcusschiesser: why do we need it on the server if we don't show it on the client?

templates/types/streaming/nextjs/app/components/ui/upload-csv-preview.tsx
```diff
@@ -1,55 +1,88 @@
...
-            src={SheetIcon}
-            alt="SheetIcon"
-          />
-        </div>
-        <div className="overflow-hidden">
-          <div className="truncate font-semibold">
-            {filename} ({fileSizeInKB} KB)
+    <Drawer direction="left">
+      <DrawerTrigger asChild>
+        <div className="p-2 w-60 max-w-60 bg-secondary rounded-lg text-sm relative cursor-pointer">
```
marcusschiesser, line 30: factor out the preview class (same file don't export)?

marcusschiesser: no need to detach, I just mean that the code inside `DrawerTrigger` could be a own component that we keep in this file

marcusschiesser (approved):

PR #116, opened 2024-06-06
refactor: use annotations instead of data

marcusschiesser:
- remove sending images and CSVs as `data` (avoids sending content twice)
- use annotations to generate `MessageContentDetail[]`
- as GPT4 ignores `MessageContentDetail` from previous messages (only pure string messages seem to be used), fallback to use annotation from last user message 

## Summary by CodeRabbit

- **Refactor**
  - Simplified the `ChatInput` component by grouping related properties into a single object.
  - Updated the `useCsv` hook to streamline state management and file handling.

- **New Features**
  - Improved file upload functionality in the chat input, making it more intuitive and efficient.

### DIFF
--- a/templates/types/streaming/express/src/controllers/llamaindex-stream.ts
+++ b/templates/types/streaming/express/src/controllers/llamaindex-stream.ts
@@ -1,4 +1,5 @@
 import {
+  JSONValue,
   StreamData,
   createCallbacksTransformer,
   createStreamDataTransformer,
@@ -7,6 +8,7 @@ import {
 } from "ai";
 import {
   MessageContent,
+  MessageContentDetail,
   Metadata,
   NodeWithScore,
   Response,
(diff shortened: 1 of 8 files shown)


templates/types/streaming/nextjs/app/components/ui/chat/chat-input.tsx
```diff
@@ -63,17 +65,12 @@ export default function ChatInput(
...
-      },
-      {
-        data: { imageUrl, csvFiles: files },
-      },
-    );
+    props.append!({
+      content: props.input,
+      role: "user",
+      createdAt: new Date(),
+      annotations: attachments,
```
marcusschiesser, line 72: @thucpn these are just the annotations of the current message right?

but the request also sends all other annotations, where is this handled?

templates/types/streaming/nextjs/app/components/ui/chat/chat-input.tsx
```diff
@@ -63,17 +65,12 @@ export default function ChatInput(
...
+    });
     setImageUrl(null);
     resetUploadedFiles();
     props.setInput!("");
```
marcusschiesser, line 76: is this not something we also want generally in onSubmit to happen (even if there's no attachment)?

templates/types/streaming/nextjs/app/components/ui/chat/chat-input.tsx
```diff
@@ -26,24 +25,27 @@ export default function ChatInput(
...
 ) {
   const [imageUrl, setImageUrl] = useState<string | null>(null);
-  const { files, uploadNew, removeFile, resetUploadedFiles } = useCsv(
-    props.messages,
-  );
+  const {
+    uploadedFiles: uploadedCsvFiles,
+    uploadNew,
+    removeFile,
+    resetUploadedFiles,
```
marcusschiesser, line 32: simplify naming?

```suggestion
    files: csvFiles,
    upload,
    remove,
    reset,
```

templates/types/streaming/nextjs/app/components/ui/chat/chat-input.tsx
marcusschiesser: e.g. can we stop re-sending the annotations that are not from user messages (like tools, etc.)

templates/types/streaming/express/src/controllers/llamaindex-stream.ts
```diff
@@ -20,50 +22,73 @@ type LlamaIndexResponse =
...
-  }
+const convertAnnotations = (
+  annotations: JSONValue[],
+): MessageContentDetail[] => {
+  const content: MessageContentDetail[] = [];
+  annotations.forEach((annotation: JSONValue) => {
+    // first skip invalid annotation
+    if (
+      !(
+        annotation &&
```
marcusschiesser, line 47: @thucpn happy for suggestions how to simplify this (without using `any` :D )

templates/types/streaming/fastapi/app/api/routers/models.py
```diff
@@ -2,19 +2,14 @@
 import logging
 from pydantic import BaseModel, Field, validator
 from pydantic.alias_generators import to_camel
-from typing import List, Any, Optional, Dict
+from typing import List, Any, Optional, Dict, Any
```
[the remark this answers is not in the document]
marcusschiesser: @leehuwuj the rabbit is right

templates/types/streaming/nextjs/app/components/ui/chat/use-csv.ts
```diff
@@ -1,83 +1,36 @@
...
-  const files = useMemo(() => {
-    return [...availableFiles, ...uploadedFiles];
-  }, [availableFiles, uploadedFiles]);
-
-  useEffect(() => {
-    const items = getAvailableCsvFiles(messages);
-    setAvailableFiles(items.map((data) => ({ ...data, type: "available" })));
-  }, [messages]);
+export function useCsv() {
+  const [uploadedFiles, setUploadedFiles] = useState<CsvFile[]>([]);
```
marcusschiesser, line 7: just files?

templates/types/streaming/fastapi/app/api/routers/models.py
```diff
@@ -80,9 +83,20 @@ def get_last_message_content(self) -> str:
         """
         Get the content of the last message along with the data content if available
         """
-        message_content = self.messages[-1].content
-        if self.data:
-            message_content += "\n" + self.data.to_raw_content()
+        if len(self.messages) == 0:
```
marcusschiesser, line 86: this mixes the concerns: of getting the last message, converting it from vercel/ai to llamaindex format and looking up the annotation to use.
my guess is that  `looking up the annotation` is something that users might want to modify and also the comment in the TS code why we do it seems helpful to me

templates/types/streaming/nextjs/app/components/ui/chat/chat-input.tsx
marcusschiesser: ah then we better leave this as an optimization step, we want the code to be as small as possible

marcusschiesser: @thucpn good job overriding the submit!

templates/types/streaming/express/src/controllers/llamaindex-stream.ts
marcusschiesser: might be a bit overkill for now, but please consider using it in a next PR if we're using more annotations

templates/types/streaming/nextjs/app/components/ui/chat/chat-input.tsx
marcusschiesser: i mean the `props.setInput!("");` - should it not be in `onSubmit`?

templates/types/streaming/fastapi/app/api/routers/models.py
marcusschiesser, line 5:
```suggestion
from typing import List, Any, Optional, Dict
```

templates/types/streaming/fastapi/app/api/routers/models.py
```diff
@@ -80,9 +83,20 @@ def get_last_message_content(self) -> str:
         """
         Get the content of the last message along with the data content if available
```
marcusschiesser, line 84:
```suggestion
        Get the content of the last message along with the data content if available. Fallback to use data content from previous messages
```

PR #127, opened 2024-06-13
Reuse function tool instance and improve e2b interpreter tool

leehuwuj:
## Summary by CodeRabbit

- **New Features**
	- Enhanced efficiency of the Python interpreter tool by reusing function instances.
	- Added a caching mechanism for parsed OpenAPI specs to improve performance.

- **Improvements**
	- Updated the description of the Python interpreter tool to emphasize secure code execution, error handling, and pip package installation.

### DIFF
--- a/helpers/tools.ts
+++ b/helpers/tools.ts
@@ -111,7 +111,7 @@ export const supportedTools: Tool[] = [
 - The python code runs in a Jupyter notebook. Every time you call the 'interpreter' tool, the python code is executed in a separate cell. 
 - You are given tasks to complete and you run python code to solve them.
 - It's okay to make multiple calls to interpreter tool. If you get an error or the result is not what you expected, you can call the tool again. Don't give up too soon!
-- Display visualizations using matplotlib or any other visualization library directly in the notebook.
+- Plot visualizations using matplotlib or any other visualization library directly in the notebook.
 - You can install any pip package (if it exists) by running a cell with pip install.`,
       },
     ],
--- a/helpers/tools.ts
+++ b/helpers/tools.ts
@@ -107,13 +107,12 @@ export const supportedTools: Tool[] = [
       {
         name: TOOL_SYSTEM_PROMPT_ENV_VAR,
         description: "System prompt for code interpreter tool.",
-        value: `You are a Python interpreter.
-        - You are given tasks to complete and you run python code to solve them.
-        - The python code runs in a Jupyter notebook. Every time you call \`interpreter\` tool, the python code is executed in a separate cell. It's okay to make multiple calls to \`interpreter\`.
-        - Display visualizations using matplotlib or any other visualization library directly in the notebook. Shouldn't save the visualizations to a file, just return the base64 encoded data.
-        - You can install any pip package (if it exists) if you need to but the usual packages for data analysis are already preinstalled.
-        - You can run any python code you want in a secure environment.
-        - Use absolute url from result to display images or any other media.`,
+        value: `-You are a Python interpreter that can run any python code in a secure environment.
+- The python code runs in a Jupyter notebook. Every time you call the 'interpreter' tool, the python code is executed in a separate cell. 
+- You are given tasks to complete and you run python code to solve them.
+- It's okay to make multiple calls to interpreter tool. If you get an error or the result is not what you expected, you can call the tool again. Don't give up too soon!
+- Display visualizations using matplotlib or any other visualization library directly in the notebook.
+- You can install any pip package (if it exists) by running a cell with pip install.`,
       },
     ],
   },
--- a/templates/components/engines/python/agent/tools/__init__.py
+++ b/templates/components/engines/python/agent/tools/__init__.py
@@ -31,7 +31,7 @@ class ToolFactory:
                 return tool_spec.to_tool_list()
             else:
                 module = importlib.import_module(f"{source_package}.{tool_name}")
-                tools = getattr(module, "get_tools")()
+                tools = module.get_tools()
                 if not all(isinstance(tool, FunctionTool) for tool in tools):
                     raise ValueError(
                         f"The module {module} does not contain valid tools"
--- a/templates/components/engines/python/agent/tools/__init__.py
+++ b/templates/components/engines/python/agent/tools/__init__.py
@@ -19,15 +19,6 @@ class ToolFactory:
         ToolType.LOCAL: "app.engine.tools",
     }
 
-    @staticmethod
-    @cached(
-        LRUCache(maxsize=100),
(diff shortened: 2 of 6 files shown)


helpers/tools.ts
```diff
@@ -107,13 +107,12 @@ export const supportedTools: Tool[] = [
       {
         name: TOOL_SYSTEM_PROMPT_ENV_VAR,
         description: "System prompt for code interpreter tool.",
-        value: `You are a Python interpreter.
-        - You are given tasks to complete and you run python code to solve them.
-        - The python code runs in a Jupyter notebook. Every time you call \`interpreter\` tool, the python code is executed in a separate cell. It's okay to make multiple calls to \`interpreter\`.
-        - Display visualizations using matplotlib or any other visualization library directly in the notebook. Shouldn't save the visualizations to a file, just return the base64 encoded data.
-        - You can install any pip package (if it exists) if you need to but the usual packages for data analysis are already preinstalled.
-        - You can run any python code you want in a secure environment.
-        - Use absolute url from result to display images or any other media.`,
```
leehuwuj, line 116: @thucpn why we need this instruction `Use absolute url from result to display images or any other media.`?

helpers/tools.ts
```diff
@@ -107,13 +107,12 @@ export const supportedTools: Tool[] = [
       {
         name: TOOL_SYSTEM_PROMPT_ENV_VAR,
         description: "System prompt for code interpreter tool.",
-        value: `You are a Python interpreter.
-        - You are given tasks to complete and you run python code to solve them.
-        - The python code runs in a Jupyter notebook. Every time you call \`interpreter\` tool, the python code is executed in a separate cell. It's okay to make multiple calls to \`interpreter\`.
-        - Display visualizations using matplotlib or any other visualization library directly in the notebook. Shouldn't save the visualizations to a file, just return the base64 encoded data.
```
leehuwuj, line 113: It doesn't need to return the chart as base64, just show it directly in the cell then the e2b will automatically encode it as png. So, we can remove `Shouldn't save the visualizations to a file, just return the base64 encoded data.`

templates/components/engines/python/agent/tools/__init__.py
```diff
@@ -19,15 +19,6 @@ class ToolFactory:
...
 
-    @staticmethod
-    @cached(
-        LRUCache(maxsize=100),
-        key=lambda tool_type, tool_name, config: (
-            tool_type,
-            tool_name,
-            json.dumps(config, sort_keys=True),
-        ),
-    )
```
leehuwuj, line 30: We added this to reuse the tool instance for the OpenAPI tool feature. But it's also applied to all the other tools too. I think we now should accept the issue and find another way to reuse the parsed openapi result later. @marcusschiesser what do you think?

templates/components/engines/python/agent/tools/__init__.py
```diff
@@ -40,7 +31,7 @@ def load_tools(tool_type: str, tool_name: str, config: dict) -> list[FunctionToo
                 return tool_spec.to_tool_list()
             else:
                 module = importlib.import_module(f"{source_package}.{tool_name}")
-                tools = getattr(module, "tools")
+                tools = getattr(module, "get_tools")()
```
leehuwuj, line 34: basically, we just change the tool singleton instance into a function to initiate the tool per request.

helpers/tools.ts
marcusschiesser: @leehuwuj then please also change this in TS

templates/components/engines/python/agent/tools/interpreter.py
```diff
@@ -101,50 +115,28 @@ def parse_result(self, result) -> List[InterpreterExtraResult]:
...
+            else:
+                results = self.parse_result(exec.results[0])
+                output = E2BToolOutput(is_error=False, logs=exec.logs, results=results)
+        return output
 
 
-# Specify as functions tools to be loaded by the ToolFactory
-tools = [FunctionTool.from_defaults(code_interpret)]
+def get_tools():
+    return [FunctionTool.from_defaults(E2BCodeInterpreter().interpret)]
```
marcusschiesser, line 142: are we're not recreating a `E2BCodeInterpreter` with each `get_tools` call?

templates/components/engines/python/agent/tools/__init__.py
marcusschiesser: how about caching it in the OpenAPI tool (see my TS code changes for OpenAPI)?

helpers/tools.ts
leehuwuj: @thucpn i think that issue due to the above instruction: `Shouldn't save the visualizations to a file, just return the base64 encoded data.` . It should be resolved now because we have already removed that.

templates/components/engines/python/agent/tools/__init__.py
marcusschiesser: @leehuwuj while doing this (as you're going to change the OpenAPI tool), you can also add the OpenAPI spec description to the tool that is retrieving the spec (see https://github.com/run-llama/create-llama/blob/3079162806017ae7de88f98a9e013066160ca111/templates/components/engines/typescript/agent/tools/openapi-action.ts#L134) - then both OpenAPI tools would behave the same in TS and Python

helpers/tools.ts
leehuwuj: The system prompt is already available for Python and TS

templates/components/engines/python/agent/tools/__init__.py
leehuwuj: > how about caching it in the OpenAPI tool (see my TS code changes for OpenAPI)?

I updated to cache the OpenAPI tool with the same mechanism as TS.  


> you can also add the OpenAPI spec description to the tool that is retrieving the spec

Unfortunately, we just simply combine the OpenAPI and Request tool, they're both ToolSpec so the tool descriptions are automatically inferenced by the function description. To add the openapi info into the tool description we must change how ToolSpec behaves which makes the code become complicated.

marcusschiesser: Got it. added a new task for this

helpers/tools.ts
marcusschiesser: @leehuwuj you also tested the new prompt with TS?

templates/components/engines/python/agent/tools/interpreter.py
marcusschiesser: ah got it once for calling `get_tools` we're creating e2b, but `interpret` is called by the LLM in the same request with the same e2b instance

marcusschiesser (approved):

PR #131, opened 2024-06-17
Starting to add Groq as a provider

Jac-Zac:
I'd like to propose adding support for Groq as a model provider. I'm not deeply familiar with the existing implementation, but I believe this could be useful. If anyone is interested in help me with it, I'd appreciate the assistance (since I really do not have much time lately). I hope this is a good starting point.

## Summary by CodeRabbit

- **New Features**
  - Introduced Groq as a new model provider, allowing users to configure models and embeddings specifically for Groq.
  - Added support for setting up Groq API keys and model configuration through user prompts.

- **Improvements**
  - Expanded model selection options to include Groq, alongside existing providers like OpenAI and Anthropic.

### DIFF
--- a/helpers/providers/groq.ts
+++ b/helpers/providers/groq.ts
@@ -10,8 +10,7 @@ const MODELS = [
 ];
 const DEFAULT_MODEL = MODELS[0];
 
-// TODO: get embedding vector dimensions from the anthropic sdk (currently not supported)
-// Use huggingface embedding models for now
+// Use huggingface embedding models for now as Groq doesn't support embedding models
 enum HuggingFaceEmbeddingModelType {
   XENOVA_ALL_MINILM_L6_V2 = "all-MiniLM-L6-v2",
   XENOVA_ALL_MPNET_BASE_V2 = "all-mpnet-base-v2",
--- a/helpers/providers/groq.ts
+++ b/helpers/providers/groq.ts
@@ -1,0 +1,104 @@
+import ciInfo from "ci-info";
+import prompts from "prompts";
+import { ModelConfigParams } from ".";
+import { questionHandlers, toChoice } from "../../questions";
+
+const MODELS = [
+  "llama3-8b",
+  "llama3-70b",
+  "mixtral-8x7b",
+];
+const DEFAULT_MODEL = MODELS[0];
+
+// TODO: get embedding vector dimensions from the anthropic sdk (currently not supported)
+// Use huggingface embedding models for now
+enum HuggingFaceEmbeddingModelType {
+  XENOVA_ALL_MINILM_L6_V2 = "all-MiniLM-L6-v2",
+  XENOVA_ALL_MPNET_BASE_V2 = "all-mpnet-base-v2",
+}
+type ModelData = {
+  dimensions: number;
+};
+const EMBEDDING_MODELS: Record<HuggingFaceEmbeddingModelType, ModelData> = {
+  [HuggingFaceEmbeddingModelType.XENOVA_ALL_MINILM_L6_V2]: {
+    dimensions: 384,
+  },
+  [HuggingFaceEmbeddingModelType.XENOVA_ALL_MPNET_BASE_V2]: {
+    dimensions: 768,
+  },
+};
+const DEFAULT_EMBEDDING_MODEL = Object.keys(EMBEDDING_MODELS)[0];
+const DEFAULT_DIMENSIONS = Object.values(EMBEDDING_MODELS)[0].dimensions;
+
+type GroqQuestionsParams = {
+  apiKey?: string;
+  askModels: boolean;
+};
+
+export async function askGroqQuestions({
+  askModels,
+  apiKey,
+}: GroqQuestionsParams): Promise<ModelConfigParams> {
+  const config: ModelConfigParams = {
+    apiKey,
+    model: DEFAULT_MODEL,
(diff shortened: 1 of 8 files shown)


templates/types/streaming/fastapi/app/settings.py
```diff
@@ -88,9 +91,30 @@ def init_azure_openai():
     Settings.embed_model = AzureOpenAIEmbedding(**embedding_config)
 
 
+def init_groq():
+    from llama_index.embeddings.huggingface import HuggingFaceEmbedding
+    from llama_index.llms.groq import Groq
+
+    model_map: Dict[str, str] = {
+        "llama3-8b": "llama3-8b-8192",
```
marcusschiesser, line 99: @Jac-Zac do we need this model map for Groq in Python? e.g. i can't use `llama3-8b` directly in Groq in Python?
Ideally, we won't have this map as in TS (see `settings.ts`)

(note: for hugging face embedding, we need the mapping)

helpers/providers/groq.ts
```diff
@@ -0,0 +1,104 @@
...
+
+const MODELS = [
+  "llama3-8b",
+  "llama3-70b",
+  "mixtral-8x7b",
+];
+const DEFAULT_MODEL = MODELS[0];
+
+// TODO: get embedding vector dimensions from the anthropic sdk (currently not supported)
+// Use huggingface embedding models for now
```
marcusschiesser, line 14:
```suggestion
// Use huggingface embedding models for now as Groq doesn't support embedding models
```

marcusschiesser (commented):
@Jac-Zac looks great, thanks a lot! just a few questions

templates/types/streaming/fastapi/app/settings.py
Jac-Zac: I think we need it since those are the models that are provided by the groq api: https://console.groq.com/docs/models

Jac-Zac: I'm not sure in which exactly, I took inspiration from the anthropic one since I think it quite similar actually. Also I have added a commit which I think might be needed, we can also squash it later

templates/types/streaming/express/src/controllers/engine/settings.ts
```diff
@@ -85,6 +90,21 @@ function initAnthropic() {
...
 
+
+function initGroq() {
+  // To review
+  const embedModelMap: Record<string, string> = {
+    "all-MiniLM-L6-v2": "Xenova/all-MiniLM-L6-v2",
+    "all-mpnet-base-v2": "Xenova/all-mpnet-base-v2",
+  };
+  Settings.llm = new Groq({
+    model: process.env.MODEL as keyof typeof ALL_AVAILABLE_GROQ_MODELS,
```
marcusschiesser, line 101: Needs modelMap lookup like in settings.py - values of the map need to be of type `keyof typeof ALL_AVAILABLE_GROQ_MODELS`
```suggestion
    model: modelMap[process.env.MODEL!],
```
(same also for nextjs)

templates/types/streaming/fastapi/app/settings.py
marcusschiesser: @Jac-Zac I added a comment where to put the lookup

templates/types/streaming/express/src/controllers/engine/settings.ts
Jac-Zac: You mean doing something like this ? ```diff --git a/templates/types/streaming/nextjs/app/api/chat/engine/settings.ts b/templates/types/streaming/nextjs/app/api/chat/engine/settings.ts
index 48b0cac..3987cc1 100644
--- a/templates/types/streaming/nextjs/app/api/chat/engine/settings.ts
+++ b/templates/types/streaming/nextjs/app/api/chat/engine/settings.ts
@@ -79,9 +79,17 @@ function initGroq() {
     "all-MiniLM-L6-v2": "Xenova/all-MiniLM-L6-v2",
     "all-mpnet-base-v2": "Xenova/all-mpnet-base-v2",
   };
+
+  const modelMap: Record<string, string> = {
+        "llama3-8b": "llama3-8b-8192",
+        "llama3-70b": "llama3-70b-8192",
+        "mixtral-8x7b": "mixtral-8x7b-32768",
+  }
+
   Settings.llm = new Groq({
-    model: process.env.MODEL as keyof typeof ALL_AVAILABLE_GROQ_MODELS,
+    model: modelMap[process.env.MODEL!],
   });
+
   Settings.embedModel = new HuggingFaceEmbedding({
     modelType: embedModelMap[process.env.EMBEDDING_MODEL!],
   });
``` Tough honestly I'm not really sure how to do those things I just though of doing things similarly to what was done for anthropic but actually I haven't really understood this first line: ```import { ALL_AVAILABLE_ANTHROPIC_MODELS } from "llamaindex/llm/anthropic";
``` I mean I have never used typescript that is way I struggle a bit with that part

marcusschiesser: @Jac-Zac yes exactly, that looks good. The `Groq` class supports `string` type as model, so don't worry about `ALL_AVAILABLE_GROQ_MODELS`.
This code should run, can you test it on TS?

marcusschiesser (approved):
@Jac-Zac thanks that looks great!

PR #150, opened 2024-07-02
feat: Add llama-agent template

leehuwuj:
## Summary by CodeRabbit

- **New Features**
  - Introduced a new FastAPI project template for multi-agent applications.
  - Added support for various model providers (OpenAI, Groq, Ollama, Anthropic, Gemini, Azure-OpenAI).
  - Implemented a control plane server and message queue setup.
  - Provided dummy and query engine agent functionalities.
  - Added task result service for managing message processing.

- **Enhancements**
  - Improved environment variable management with `load_from_env` function.
  - Extended template types to include "multiagent."
  - Updated dependencies and project configurations for multi-agent support.

- **Documentation**
  - Added comprehensive README for the new multi-agent FastAPI project template.


create-app.ts
```diff
@@ -152,6 +152,7 @@ export async function createApp({
   }
 
   if (
+    template === "streaming" &&
```
marcusschiesser, line 155: Why is that not true for Multi-Agent?

leehuwuj: ah, looks like i hacked this to by pass an error, but it's great to add this for multi-agent app

questions.ts
```diff
@@ -395,6 +401,10 @@ export const askQuestions = async (
     return; // early return - no further questions needed for llamapack projects
   }
 
+  if (program.template === "multiagents") {
+    // TODO: multi-agents currently only supports FastAPI
```
marcusschiesser, line 405: TODO is not for this PR

templates/types/multi-agents/fastapi/README-template.md
```diff
@@ -0,0 +1,49 @@
...
+- Message queue (port 8000): To exchange the message between services
+- Control plane (port 8001): A gateway to manage the tasks and services.
+- Human consumer (port 8002): To handle result when the task is completed.
+- Agent services (from port 8003 onwards): Default is a `query_engine` agent which helping on query information from the index to answer the question.
+
+The ports listed above are set by default, but you can change them in the `.env` file.
+
+## Learn More
+
+To learn more about LlamaIndex, take a look at the following resources:
```
marcusschiesser, line 45: add llama-agents here too

templates/types/multi-agents/fastapi/README-template.md
```diff
@@ -0,0 +1,49 @@
...
+```shell
+poetry run llama-agents monitor --control-plane-url http://127.0.0.1:8001
+```
+
+## Services:
+
+- Message queue (port 8000): To exchange the message between services
+- Control plane (port 8001): A gateway to manage the tasks and services.
+- Human consumer (port 8002): To handle result when the task is completed.
+- Agent services (from port 8003 onwards): Default is a `query_engine` agent which helping on query information from the index to answer the question.
```
marcusschiesser, line 39:
```suggestion
- Agent service `query_engine` (port 8003): Agent that can query information from the configured LlamaIndex index.
- Agent service `dummy_agent` (port 8004): A dummy agent that does nothing. Good starting point to add more agents. 
```

questions.ts
```diff
@@ -694,7 +707,8 @@ export const askQuestions = async (
     }
   }
 
-  if (!program.tools) {
+  if (!program.tools && program.template === "streaming") {
+    // TODO: allow to select tools also for multi-agent framework
```
marcusschiesser, line 711: TODO is not for this PR

templates/types/multi-agents/fastapi/app/settings.py
```diff
@@ -0,0 +1,146 @@
+import os
```
marcusschiesser, line 1: TODO: also copy settings.py from components (reuse it in streaming and this new template)
Provide a dummy in both templates that does nothing

templates/types/multi-agents/fastapi/env.example
```diff
@@ -0,0 +1,37 @@
...
+# Specify the port for the multi-agent services
+# Core services
+MESSAGE_QUEUE_PORT=8000
+CONTROL_PLANE_PORT=8001
+HUMAN_CONSUMER_PORT=8002
+# Agent services
+AGENT_QUERY_ENGINE_PORT=8003
+AGENT_QUERY_ENGINE_DESCRIPTION="Can query the knowledge about US letter standards"
+AGENT_DUMMY_PORT=8004
+AGENT_DUMMY_DESCRIPTION="Does nothing"
```
marcusschiesser, line 37: how about the following:

1. use defaults for the ports in code (then configuration via env variables is optional)
2. inline `AGENT_DUMMY_DESCRIPTION` (no need for an env variable)
3. rename `AGENT_QUERY_ENGINE_DESCRIPTION` to `QUERY_ENGINE_DESCRIPTION`, generate a default in env-variable.ts and also use it for the single-agent's `QueryEngineTool` (Python and TS)

From my understanding, with these changes, we could use the generated .env file

helpers/multi-agents.ts
leehuwuj, line 1: Why you removed this? We still need copy for the multi-agent templates right?

leehuwuj: Ah, i guess that's todo task

templates/types/multi-agents/fastapi/env.example
leehuwuj: Agree with 1. and 2.
About 3. I think we can keep the `AGENT_QUERY_ENGINE_DESCRIPTION`, the reason is the agent description is a bit different with the tool or the index and one agent can use multiple tools/index too. We should work on the tool dynamic description and index description first then we will able to construct the agent from those metadata.

templates/types/multi-agents/fastapi/app/settings.py
leehuwuj: I updated for copying the settings.py. But, why you need the dummy agent in streaming template?

questions.ts
```diff
@@ -122,6 +122,7 @@ const getVectorDbChoices = (framework: TemplateFramework) => {
 export const getDataSourceChoices = (
   framework: TemplateFramework,
   selectedDataSource: TemplateDataSource[],
+  template?: string,
```
marcusschiesser, line 125:
```suggestion
  template?: TemplateType,
```

templates/types/multi-agents/fastapi/app/settings.py
marcusschiesser: hehe, i didn't mean the dummy agent. i was talking about a dummy `init_settings` function (that way the the template would still run without copying) - we're using this pattern in TS templates. but thinking about it, you're right, we shouldn't do it

PR #153, opened 2024-07-05
feat: able to display file url from llamacloud

thucpn:
## Summary by CodeRabbit

- **New Features**
  - Added support for displaying files sourced from LlamaCloud indexes.
  - Introduced directory creation for "llamacloud" within the "output" directory.
  
- **Enhancements**
  - Improved document processing and metadata handling for files.
  - Enhanced chat message content handling and document ID management.
  - Added functionality for loading and saving documents based on MIME types.

- **Bug Fixes**
  - Improved error handling during asynchronous operations.


templates/types/streaming/fastapi/app/api/routers/models.py
```diff
@@ -112,6 +113,40 @@ def is_last_message_from_user(self) -> bool:
         return self.messages[-1].role == MessageRole.USER
 
 
+class LLamaCloudFileService(BaseModel):
```
marcusschiesser, line 116: move to own file?

leehuwuj: This not look like a service integration rather than data model to me, better to move the code to other place and no need to use the `BaseModel` here

templates/types/streaming/fastapi/app/api/routers/models.py
```diff
@@ -124,6 +126,12 @@ def from_source_node(cls, source_node: NodeWithScore):
         metadata = source_node.node.metadata
         url = metadata.get("URL")
 
+        # if metadata has pipeline_id, get file url from LLamaCloudFileService
+        pipeline_id = metadata.get("pipeline_id")
+        if pipeline_id:
+            file_name = metadata.get("file_name")
+            url = LLamaCloudFileService.get_file_url(file_name, pipeline_id)
```
leehuwuj, line 133: This fetches data from the llama cloud service, which means it can have network errors. How about we add error handling here that accepts the error and uses the null URL?

templates/types/streaming/fastapi/app/api/routers/models.py
```diff
@@ -124,6 +126,12 @@ def from_source_node(cls, source_node: NodeWithScore):
...
+
```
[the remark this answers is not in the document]
leehuwuj: Yes, the suggestion from @coderrabit makes sense

templates/types/streaming/fastapi/app/api/routers/models.py
marcusschiesser: good idea

templates/types/streaming/nextjs/app/components/ui/chat/chat-message/chat-sources.tsx
```diff
@@ -54,34 +56,66 @@ export function ChatSources({ data }: { data: SourceData }) {
...
+        console.error("Failed to fetch PDF file", error);
+      }
+    };
+    if (nodeInfo.url && nodeInfo.filename?.endsWith(".pdf")) {
+      checkPdfUrl(nodeInfo.url);
+    }
+  }, [nodeInfo.filename, nodeInfo.url]);
+
+  if (isValidPdf) {
+    return (
```
marcusschiesser, line 96:
```suggestion
    // only use the PDF Viewer if we're dealing with a PDF that the client can fetch
    // (e.g. LlamaCloud's PDFs are not fetchable due to CORS restrictions)
    return (
```

templates/types/streaming/nextjs/app/components/ui/chat/chat-message/chat-sources.tsx
```diff
@@ -23,6 +23,7 @@ function SourceNumberButton({ index }: { index: number }) {
 type NodeInfo = {
   id: string;
   url?: string;
+  filename?: string;
```
marcusschiesser, line 26: why do we need filename?

templates/types/streaming/nextjs/app/api/chat/stream-helper.ts
```diff
@@ -23,20 +32,23 @@ function getNodeUrl(metadata: Metadata) {
...
-export function appendSourceData(
+export async function appendSourceData(
   data: StreamData,
   sourceNodes?: NodeWithScore<Metadata>[],
 ) {
   if (!sourceNodes?.length) return;
+  const nodes = await Promise.all(
+    sourceNodes.map(async (node) => ({
+      ...node.node.toMutableJSON(),
+      id: node.node.id_,
```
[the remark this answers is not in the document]
marcusschiesser: ah sorry, i read that wrong

marcusschiesser (approved):

marcusschiesser:
TS code looks good - @leehuwuj please check the python code and then you can merge it

templates/types/streaming/fastapi/app/api/routers/chat.py
```diff
@@ -41,6 +74,13 @@ async def _chat_response_generator():
                 response = await chat_engine.astream_chat(
                     last_message_content, messages
                 )
+                source_nodes = [
+                    SourceNodes.from_source_node(node)
+                    for node in response.source_nodes
+                ]
+                process_source_nodes(source_nodes, background_tasks)
```
leehuwuj, line 81: We should move this to callback event handler, but we'll need a way to pass the injected `background_tasks` object to the handler.

templates/types/streaming/fastapi/app/tasks/llama_cloud.py
```diff
@@ -0,0 +1,89 @@
...
+        # Download the file
+        with requests.get(url, stream=True) as r:
+            r.raise_for_status()
+            with open(local_file_path, "wb") as f:
+                for chunk in r.iter_content(chunk_size=8192):
+                    f.write(chunk)
+        logger.info("File downloaded successfully")
+
+    @classmethod
+    def task_download_llamacloud_pipeline_file(
```
marcusschiesser, line 42: doesn't seem to have a dependency to `BackgroundTasks` so we could also call this as a normal method right? so basically, we can see this as a service method

templates/types/streaming/fastapi/app/tasks/llama_cloud.py
```diff
@@ -0,0 +1,89 @@
+import os
+import logging
+import requests
+from typing import List, Any, Dict, Optional
+
+
+logger = logging.getLogger("uvicorn")
+
+
+class LLamaCloudFile:
```
marcusschiesser, line 10: only two methods are public, can we mark the private methods, so it's clear?
(same for the other file service)

templates/types/streaming/fastapi/app/api/routers/chat.py
```diff
@@ -15,17 +17,48 @@
...
 
 
+def process_source_nodes(
+    source_nodes: List[SourceNodes],
+    background_tasks: BackgroundTasks,
+):
+    """
+    Start background tasks on the source nodes if needed.
+    """
+    to_download_files = [
```
marcusschiesser, line 40:
```suggestion
    files_to_download = [
```

templates/types/streaming/fastapi/app/api/routers/chat.py
marcusschiesser: i think the event handler should not deal with this concern. better pass a processor function to the event handler that can deal with `source_nodes` outside of the event handler code (here downloading the llamafiles)

templates/components/llamaindex/typescript/streaming/events.ts
```diff
@@ -6,23 +6,53 @@ import {
   ToolCall,
   ToolOutput,
 } from "llamaindex";
+import { LLamaCloudFileService } from "./service";
 
-export function appendSourceData(
+async function getNodeUrl(metadata: Metadata) {
```
marcusschiesser, line 11: @thucpn this function is now at the bottom of the file (as it's not exported) - merge it there please

templates/components/llamaindex/typescript/streaming/events.ts
```diff
@@ -113,3 +121,25 @@ function getNodeUrl(metadata: Metadata) {
   }
   return undefined;
 }
+
+async function getNodeUrl(metadata: Metadata) {
```
marcusschiesser, line 125: @thucpn you still have the old `getNodeUrl` function

templates/components/llamaindex/typescript/streaming/service.ts
```diff
@@ -0,0 +1,98 @@
...
+      }
+      return null;
+    } catch (error) {
+      console.error("Error fetching file from LlamaCloud:", error);
+      return null;
+    }
+  }
+
+  static downloadFile(url: string, fileId: string, filename: string) {
+    const directory = "data/private"; // TODO: move to output/llamacloud later
```
marcusschiesser, line 60: move to output/llamacloud

marcusschiesser: use const

templates/types/streaming/fastapi/app/api/routers/chat.py
```diff
@@ -1,8 +1,15 @@
 import os
 import logging
 
+from typing import List
 from aiostream import stream
+<<<<<<< HEAD
```
marcusschiesser, line 6: ooops

templates/types/streaming/fastapi/app/api/controllers/llama_cloud.py
```diff
@@ -0,0 +1,88 @@
...
+
+class LLamaCloudFileController:
+
+    LLAMA_CLOUD_URL = "https://cloud.llamaindex.ai/api/v1"
+    LOCAL_STORE_PATH = "output/llamacloud"
+
+    DOWNLOAD_FILE_NAME_TPL = "{pipeline_id}${filename}"
+
+    @classmethod
+    def get_files(cls, pipeline_id: str) -> List[Dict[str, Any]]:
```
marcusschiesser, line 18: mark private methods?

templates/components/llamaindex/typescript/documents/helper.ts
```diff
@@ -0,0 +1,54 @@
...
+    default:
+      throw new Error(`Unsupported document type: ${mimeType}`);
+  }
+}
+
+export async function saveDocument(fileBuffer: Buffer, mimeType: string) {
+  const fileExt = MIME_TYPE_TO_EXT[mimeType];
+  if (!fileExt) throw new Error(`Unsupported document type: ${mimeType}`);
+
+  const folder = "output/uploaded";
```
marcusschiesser, line 38: const in header of file

helpers/typescript.ts
```diff
@@ -115,9 +115,37 @@ export const installTSTemplate = async ({
...
+          "llamacloud",
+        ),
+      },
+    );
   } else {
     console.log("\nUsing vector DB:", vectorDb ?? "none");
   }
+
+  // remove llamacloud folder in documents
+  await fs.rm(
```
marcusschiesser, line 138: if we move the llamacloud pipeline to a own components folder (e.g. `components/llamacloud/typescript`), then we don't need to do this, right?

templates/components/llamaindex/typescript/streaming/service.ts
```diff
@@ -0,0 +1,101 @@
...
+
+export interface LlamaCloudFile {
+  name: string;
+  file_id: string;
+  project_id: string;
+}
+
+export class LLamaCloudFileService {
+  static async getFiles(pipelineId: string): Promise<LlamaCloudFile[]> {
+    const url = `https://cloud.llamaindex.ai/api/v1/pipelines/${pipelineId}/files`;
```
marcusschiesser, line 13: use const
```suggestion
    const url = `${LLAMA_CLOUD_BASE_URL}/pipelines/${pipelineId}/files`;
```

templates/components/llamaindex/typescript/streaming/service.ts
```diff
@@ -0,0 +1,101 @@
...
+      }
+      return null;
+    } catch (error) {
+      console.error("Error fetching file from LlamaCloud:", error);
+      return null;
+    }
+  }
+
+  static downloadFile(url: string, fileId: string, filename: string) {
+    const LLAMACKOUD_OUTPUT_DIR = "output/llamacloud";
```
marcusschiesser, line 60: and move to file header
```suggestion
    const LLAMA_CLOUD_OUTPUT_DIR = "output/llamacloud";
```

templates/components/llamaindex/typescript/streaming/service.ts
```diff
@@ -0,0 +1,101 @@
...
+    const FILE_DELIMITER = "$"; // delimiter between fileId and filename
+    const downloadedFileName = `${fileId}${FILE_DELIMITER}${filename}`;
+    const downloadedFilePath = path.join(
+      LLAMACKOUD_OUTPUT_DIR,
+      downloadedFileName,
+    );
+    const urlPrefix = `${process.env.FILESERVER_URL_PREFIX}/${LLAMACKOUD_OUTPUT_DIR}`;
```
marcusschiesser, line 67: how about reusing the existing method to download a file from the file server?

helpers/typescript.ts
```diff
@@ -115,9 +115,37 @@ export const installTSTemplate = async ({
     console.log(
       `\nUsing managed index from LlamaCloud. Ensure the ${yellow("LLAMA_CLOUD_* environment variables are set correctly.")}`,
     );
+    // copy llamacloud pipeline code for documents
+    await copy(
```
marcusschiesser, line 119: alternatively, we could not generate LlamaCloud specific code and check at runtime if we're having LlamaCloud configured (and then run the pipeline).
@leehuwuj what are your thoughts?
please make sure that it's the same behavior in the python code

templates/types/streaming/fastapi/app/api/routers/chat.py
```diff
@@ -1,32 +1,66 @@
...
+        )
+    ]
+    # Remove duplicates
+    files_to_download = list({v["pipeline_id"]: v for v in files_to_download}.values())
+    for file in files_to_download:
+        background_tasks.add_task(
+            LLamaCloudFileController.download_llamacloud_pipeline_file,
+            file_name=file.get("file_name"),
+            pipeline_id=file.get("pipeline_id"),
+        )
```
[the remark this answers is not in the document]
leehuwuj: Noted that! Yes, it's good idea

templates/components/llamaindex/typescript/documents/llamacloud/pipeline.ts
```diff
@@ -0,0 +1,16 @@
...
+import { getDataSource } from "../../engine";
+
+export async function runPipeline(documents: Document[], filename: string) {
+  const currentIndex = await getDataSource();
+  for (const document of documents) {
+    document.metadata = {
+      ...document.metadata,
+      file_name: filename,
+      private: true, // mark document as private
+      use_local_url: true, // mark use local url to display the document
```
leehuwuj, line 11: @thucpn There's no code for this, so why did we add it?
if we will use this one, i'd like to call it as `use_remote_file` - the python is also using it to construct the file url correctly.

helpers/typescript.ts
leehuwuj: in Python i also run the different pipeline based on the index type, so i'd put the code in the same file. We then no need to add this copy logic.

helpers/typescript.ts
leehuwuj: agree, I updated this

templates/types/streaming/fastapi/app/api/routers/models.py
```diff
@@ -145,44 +146,90 @@ def get_chat_document_ids(self) -> List[str]:
...
+        # Construct file url through file server
+        if url is not None or file_name is not None:
             url_prefix = os.getenv("FILESERVER_URL_PREFIX")
             if not url_prefix:
                 logger.warning(
                     "Warning: FILESERVER_URL_PREFIX not set in environment variables"
                 )
             if file_name and url_prefix:
-                if is_private:
+                if use_remote_file:
```
marcusschiesser, line 196: why not `is_llamacloud_file`?

templates/types/streaming/fastapi/app/api/routers/models.py
```diff
@@ -145,44 +146,90 @@ def get_chat_document_ids(self) -> List[str]:
...
         metadata = source_node.node.metadata
         url = metadata.get("URL")
-
-        if not url:
-            file_name = metadata.get("file_name")
-            is_private = metadata.get("private", "false") == "true"
+        pipeline_id = metadata.get("pipeline_id")
+        file_name = metadata.get("file_name")
+        is_private = metadata.get("private", "false") == "true"
+        use_remote_file = metadata.get("use_remote_file", False)
```
marcusschiesser, line 181: when is this true and it's not a llamacloud file?

templates/types/streaming/fastapi/app/api/routers/chat.py
```diff
@@ -1,32 +1,52 @@
...
+    """
+    Start background tasks on the source nodes if needed.
+    """
+    files_to_download = SourceNodes.get_download_files(nodes)
+    for file in files_to_download:
+        background_tasks.add_task(
+            LLamaCloudFileController.download_llamacloud_pipeline_file,
+            file_name=file.get("file_name"),
+            pipeline_id=file.get("pipeline_id"),
+        )
```
marcusschiesser, line 40: isn't it `file` as paramter?
```suggestion
        background_tasks.add_task(
            LLamaCloudFileController.download_llamacloud_pipeline_file,
            file
        )
```

templates/types/streaming/fastapi/app/api/controllers/llama_cloud.py
```diff
@@ -0,0 +1,88 @@
...
+        url: str, data=None, headers: Optional[Dict] = None, method: str = "get"
+    ):
+        if headers is None:
+            headers = {
+                "Accept": "application/json",
+                "Authorization": f'Bearer {os.getenv("LLAMA_CLOUD_API_KEY")}',
+            }
+        response = requests.request(method, url, headers=headers, data=data)
+        response.raise_for_status()
+        return response.json()
```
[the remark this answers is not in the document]
marcusschiesser: @leehuwuj i guess `response.raise_for_status()` is doing this?

templates/types/streaming/fastapi/app/api/controllers/llama_cloud.py
```diff
@@ -0,0 +1,88 @@
...
+import os
+from typing import Any, Dict, List, Optional
+
+import requests
+from app.api.routers.models import LlamaCloudFile
+
+logger = logging.getLogger("uvicorn")
+
+
+class LLamaCloudFileController:
```
marcusschiesser, line 11: @leehuwuj still looks like a stateless service class to me, so why not `LLamaCloudFileService`?

templates/types/streaming/fastapi/app/api/controllers/file.py
```diff
@@ -1,45 +1,36 @@
...
-    return default_meta
 
 def file_loaders_map():
     default_loaders = get_file_loaders_map()
     default_loaders[".txt"] = FlatReader
     return default_loaders
 
 
-
 class FileController:
```
marcusschiesser, line 25: `PrivateFileService`?

templates/types/streaming/fastapi/app/api/controllers/llama_cloud.py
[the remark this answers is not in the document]
leehuwuj: It'll throw an error and show the traceback, so we don't need to add a try-catch to log it here

templates/types/streaming/fastapi/app/api/controllers/llama_cloud.py
leehuwuj: It just class static method, no need to set as private

templates/components/vectordbs/python/none/generate.py
```diff
@@ -24,6 +24,7 @@ def generate_datasource():
     # Set private=false to mark the document as public (required for filtering)
     for doc in documents:
         doc.metadata["private"] = "false"
```
marcusschiesser, line 26: can't we remove this? as it's not set in LlamaCloud either

templates/types/streaming/fastapi/app/engine/generate.py
```diff
@@ -67,6 +67,7 @@ def generate_datasource():
     # Set the metadata for the documents
     for doc in documents:
         doc.metadata["private"] = "false"
```
marcusschiesser, line 69: can't we remove this?

templates/types/streaming/fastapi/app/api/routers/models.py
```diff
@@ -167,45 +167,39 @@ class SourceNodes(BaseModel):
...
     url: Optional[str]
-    use_remote_file: bool = Field(
-        default=False,
-    )
 
     @classmethod
     def from_source_node(cls, source_node: NodeWithScore):
         metadata = source_node.node.metadata
-        url = metadata.get("URL")
+
```
marcusschiesser, line 174: don't we need this? e.g. for website nodes?

templates/types/streaming/fastapi/app/api/routers/models.py
```diff
@@ -167,45 +167,39 @@ class SourceNodes(BaseModel):
...
+        if not url_prefix:
+            logger.warning(
+                "Warning: FILESERVER_URL_PREFIX not set in environment variables"
+            )
+        if file_name and url_prefix:
+            if not is_local_file:
+                if pipeline_id is None:
+                    logger.warning(
+                        "Warning: The file source is llamacloud but pipeline_id is not set. Cannot construct file url"
+                    )
```
marcusschiesser, line 189: 1. when does this case happen?
2. if it would not happen, we can remove `is_local_file`, correct?

templates/types/streaming/fastapi/app/api/controllers/llama_cloud.py
marcusschiesser: isn't python usually marking `_` prefix for private stuff? (see https://peps.python.org/pep-0008/#method-names-and-instance-variables)

I know it's not really private in Python, but that way we mark what should be called and what not

templates/types/streaming/fastapi/app/engine/generate.py
leehuwuj: If we stick with nin filter then removing this is fine

templates/types/streaming/fastapi/app/api/routers/models.py
leehuwuj: 1. I don't think this would have happened if the data was consistent. But if it have we should not construct url and just return a null url to the FE and show a warning log in BE instead.
2. `is_local_file` mean that the file was stored locally or remotely (llama cloud), we need a flag to identify where the file is. Please let me know if you have better approach for this

marcusschiesser (commented):
a

templates/components/llamaindex/typescript/streaming/events.ts
```diff
@@ -97,19 +105,27 @@ export function createCallbackManager(stream: StreamData) {
...
-  if (fileName) {
-    const folder = metadata["private"] ? "output/uploaded" : "data";
+  const fileName = metadata["file_name"];
+  if (fileName && process.env.FILESERVER_URL_PREFIX) {
+    // file_name exists and file server is configured
+    const isLocalFile = metadata["is_local_file"] === "true";
+    const pipelineId = metadata["pipeline_id"];
+    if (pipelineId && !isLocalFile) {
+      // file is from LlamaCloud and was not ingested locally
+      // TODO trigger but don't await file download and just use convention to generate the URL (see Python code)
```
marcusschiesser, line 121: @thucpn see TODO

marcusschiesser (approved):

PR #160, opened 2024-07-10
fix: add azure api version

marcusschiesser:
## Summary by CodeRabbit

- **New Features**
  - Updated the integration with Azure OpenAI to use the new API version `2024-06-01` for improved compatibility and performance.

### DIFF
--- a/templates/components/settings/python/settings.py
+++ b/templates/components/settings/python/settings.py
@@ -85,6 +85,7 @@ def init_azure_openai():
         "model": os.getenv("MODEL"),
         "temperature": float(os.getenv("LLM_TEMPERATURE", DEFAULT_TEMPERATURE)),
         "max_tokens": int(max_tokens) if max_tokens is not None else None,
+        "api_version": os.getenv("AZURE_OPENAI_API_VERSION", "2024-06-01"),
     }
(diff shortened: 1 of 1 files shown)


templates/components/settings/python/settings.py
```diff
@@ -85,6 +85,7 @@ def init_azure_openai():
         "model": os.getenv("MODEL"),
         "temperature": float(os.getenv("LLM_TEMPERATURE", DEFAULT_TEMPERATURE)),
         "max_tokens": int(max_tokens) if max_tokens is not None else None,
+        "api_version": os.getenv("AZURE_OPENAI_API_VERSION", "2024-06-01"),
```
leehuwuj, line 88: We can set it explicitly by specifying api_version but can also just use the `OPENAI_API_VERSION` variable, see this: https://github.com/run-llama/llama_index/blob/6b0fe053c381a4a7aabb46d1f42b2f863f0405d5/llama-index-integrations/llms/llama-index-llms-azure-openai/llama_index/llms/azure_openai/base.py#L34

marcusschiesser: i see, then we should anyhow get an exception if we don't set the value. then we can close this PR.

PR #189, opened 2024-07-26
Fix starter questions not working in python backend

Arputikos:
When python backend is created, /config endpoint will not return conversation starters due to pydantic starter_questions case not converted to camel properly by alias_generator

## Summary by CodeRabbit

- **New Features**
	- Introduced a `serialization_alias` for the `starter_questions` field, allowing it to be serialized as `starterQuestions` for improved API compatibility.
	- Added a patch to ensure starter questions are processed correctly in the backend, enhancing functionality and user experience.

- **Changes**
	- Removed the automatic alias generator for clearer and more manual control over serialization naming conventions.

### DIFF
--- a/.changeset/stale-maps-own.md
+++ b/.changeset/stale-maps-own.md
@@ -1,0 +1,5 @@
+---
+"create-llama": patch
+---
+
+Fix starter questions not working in python backend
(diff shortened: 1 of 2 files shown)


leehuwuj (approved):
@Arputikos LGTM! Thank you for the contribution!

PR #218, opened 2024-08-07
Groq new models

Jac-Zac:
Adding support for the latest models from Meta that are available through the groq API

## Summary by CodeRabbit

- **New Features**
  - Updated model identifiers to the latest versions, enhancing model capabilities.
  - Introduced a new tool-use preview model for improved functionality.
  - Simplified the model selection process by filtering out specific model types, focusing on LLM models.

- **Bug Fixes**
  - Corrected mappings for existing models to align with updated naming conventions.

### DIFF
--- a/templates/components/settings/python/settings.py
+++ b/templates/components/settings/python/settings.py
@@ -127,8 +127,9 @@ def init_groq():
     from llama_index.llms.groq import Groq
 
     model_map: Dict[str, str] = {
-        "llama3-8b": "llama3-8b-8192",
-        "llama3-70b": "llama3-70b-8192",
+        "llama3.1-8b": "llama-3.1-8b-instant",
+        "llama3.1-70b": "llama-3.1-70b-versatile",
+        "llama3 tool-70b": "llama3-groq-70b-8192-tool-use-preview",
         "mixtral-8x7b": "mixtral-8x7b-32768",
     }
 
--- a/helpers/providers/groq.ts
+++ b/helpers/providers/groq.ts
@@ -3,6 +3,106 @@ import prompts from "prompts";
 import { ModelConfigParams } from ".";
 import { questionHandlers, toChoice } from "../../questions";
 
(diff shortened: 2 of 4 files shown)


marcusschiesser (approved):

templates/components/settings/python/settings.py
```diff
@@ -127,8 +127,9 @@ def init_groq():
     from llama_index.llms.groq import Groq
 
     model_map: Dict[str, str] = {
-        "llama3-8b": "llama3-8b-8192",
-        "llama3-70b": "llama3-70b-8192",
+        "llama3.1-8b": "llama-3.1-8b-instant",
```
marcusschiesser, line 130: why deleting llama3?

Jac-Zac: Ah I mean no real reason just because there are the 3.1 of the same size but newer. But I mean for back compatibility we can definitely keep the old lama 3 models

Jac-Zac: I can add it back if you want.

marcusschiesser: @Jac-Zac sorry for the late reply, how about removing this mapping at all and instead retrieve the full model name in https://github.com/run-llama/create-llama/blob/main/helpers/providers/groq.ts
(by using a function similar to https://github.com/run-llama/create-llama/blob/8ce4a8513d9889427f80a7b5d1ce7bc04f35932f/helpers/providers/openai.ts#L89C1-L138C2)
endpoint to list the groq models is https://api.groq.com/openai/v1/models (see https://console.groq.com/docs/models)

Jac-Zac: I was looking at doing something like this but lately, I do not have much time sorry.

marcusschiesser:
Thanks @Jac-Zac we will finish this in https://github.com/run-llama/create-llama/pull/278

PR #237, opened 2024-08-15
fix: optional params in file upload API

thucpn:
## Summary by CodeRabbit

- **New Features**
	- The `FileUploadRequest` class now allows for easier instantiation by defaulting the `params` attribute to `None`, simplifying the usage of this class for end-users. 
	- Enhanced reliability of file uploads with LlamaCloud, ensuring a smoother and more secure experience for users when uploading private files.

### DIFF
--- a/templates/types/streaming/fastapi/app/api/routers/upload.py
+++ b/templates/types/streaming/fastapi/app/api/routers/upload.py
@@ -14,7 +14,7 @@ logger = logging.getLogger("uvicorn")
 class FileUploadRequest(BaseModel):
     base64: str
     filename: str
-    params: Any
+    params: Any = None
 
 
 @r.post("")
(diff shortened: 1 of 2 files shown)


templates/types/streaming/fastapi/app/api/routers/upload.py
```diff
@@ -14,7 +14,7 @@
 class FileUploadRequest(BaseModel):
     base64: str
     filename: str
-    params: Any
+    params: Any = None
```
leehuwuj, line 17: Just a nit! As this is a model for FastAPI endpoint so we can make it better with:
```suggestion
    params: Optional[Any]= None
```
And add import:  `from typing import Optional` . 
Or we can even describe it explicitly

```suggestion
    params: Optional[Dict] = Field(default=None, description="What does the params mean")
```

Issue #244, opened 2024-08-19
Pydantic Issue when running Ollama + FastAPI backend


BastianSpatz:
When using ollama as a model source, i get the error:

`ERROR:    Error when generating next question: 1 validation error for LLMStructuredPredictEndEvent
output
  value is not a valid dict (type=type_error.dict)`

when it wants to generate the NextQuestions. 

https://github.com/run-llama/create-llama/blob/1d93775f043cd57852413a6cb7ee6f5302fc0093/templates/types/streaming/fastapi/app/api/services/suggestion.py#L50-L55

I think this is a llama-index/pydantic problem when calling `astructured_predict` in the call `dispatcher.event(LLMStructuredPredictEndEvent(output=result))`.

Has anybody anybody seen or fixed this error?

marcusschiesser:
1. The `astructured_predict` call requires good function calling. What model are you using?
2. The code generated by create llama is only trying this call and shouldn't show the next questions if it fails - is this your behavior?

BastianSpatz:
Thanks for the reply.

1. I'm using llama 3.1 8B
2. ANd yes it just throws the error and doesnt generate questions. But the app works fine nonetheless

marcusschiesser:
@BastianSpatz 
I guess that the model is not capable enough to use `structured_predict`.

As Typescript doesn't have `structured_predict`, it's using a simple LLM call that is parsed; see:
https://github.com/run-llama/create-llama/blob/8ce4a8513d9889427f80a7b5d1ce7bc04f35932f/templates/components/llamaindex/typescript/streaming/suggestion.ts#L16-L39

Can you try using the NextJS template first with your Ollama model - if that works you could modify `suggestion.py` accordingly.

marcusschiesser:
Great. can you let me know the result, we can keep the ticket open till then

BastianSpatz:
Using the same approach as in the Typescript version it works

marcusschiesser:
cool. can you send a PR or post here your changes?

BastianSpatz:
Sorry here is what i changed in the `suggestions.py`:

``` 
NEXT_QUESTIONS_SUGGESTION_PROMPT = PromptTemplate(
    "You're a helpful assistant! Your task is to suggest the next question that user might ask. "
    "\nHere is the conversation history"
    "\n---------------------\n{conversation}\n---------------------"
    "Given the conversation history, please give me {number_of_questions} questions that you might ask next!"
[...]

PR #275, opened 2024-09-06
feat: Make suggest next questions configurable

leehuwuj:
Fix: https://github.com/run-llama/create-llama/issues/244

## Summary by CodeRabbit

- **New Features**
	- Introduced a configuration for managing the "next questions" feature, enhancing flexibility in user interactions.
	- Added functionality to suggest follow-up questions based on conversation history, improving user engagement.
	- Implemented a service for processing and storing private files, facilitating efficient data management.

- **Bug Fixes**
	- Improved response generation logic to conditionally yield suggested questions based on user settings.

- **Documentation**
	- Updated project dependencies to include `pydantic-settings` for enhanced configuration management.


helpers/env-variables.ts
```diff
@@ -487,33 +487,28 @@ It\\'s cute animal.
 };
 
 const getTemplateEnvs = (template?: TemplateType): EnvVar[] => {
-  if (template === "multiagent") {
```
marcusschiesser, line 490: ah good catch!

helpers/env-variables.ts
```diff
@@ -487,33 +487,28 @@ It\\'s cute animal.
...
-      {
-        name: "AGENT_DUMMY_PORT",
-        value: "8004",
-      },
-    ];
-  } else {
-    return [];
+  const nextQuestionEnvs: EnvVar[] = [
+    {
+      name: "NEXT_QUESTION_ENABLE",
```
marcusschiesser, line 492: don't we have the convention for citation system prompt that if it's empty we don't use it? should be consistent

templates/components/services/python/file.py
marcusschiesser, line 1: also remove from templates

templates/components/services/python/suggestion.py
marcusschiesser, line 1: also remove from templates

templates/components/services/python/suggestion.py
```diff
@@ -0,0 +1,73 @@
...
+from pydantic_settings import BaseSettings, SettingsConfigDict
+
+from app.api.routers.models import Message
+
+logger = logging.getLogger("uvicorn")
+
+
+class NextQuestionSettings(BaseSettings):
+    enable: bool = True
+    prompt_template: str = (
```
marcusschiesser, line 16: not needed if we have it in env-variables.ts

templates/components/services/python/suggestion.py
```diff
@@ -0,0 +1,73 @@
+import logging
+from typing import List, Optional
+
+from llama_index.core.prompts import PromptTemplate
+from llama_index.core.settings import Settings
+from pydantic import BaseModel
+from pydantic_settings import BaseSettings, SettingsConfigDict
```
marcusschiesser, line 7: might be confusing to add pydantic_settings just for one case?

templates/components/services/python/suggestion.py
```diff
@@ -0,0 +1,73 @@
...
+            for message in reversed(messages):
+                if message.role == "user":
+                    last_user_message = f"User: {message.content}"
+                elif message.role == "assistant":
+                    last_assistant_message = f"Assistant: {message.content}"
+                if last_user_message and last_assistant_message:
+                    break
+            conversation: str = f"{last_user_message}\n{last_assistant_message}"
+
+            output: NextQuestions = await Settings.llm.astructured_predict(
```
marcusschiesser, line 64: `Settings.llm.astructured_predict` is not always working - see https://github.com/run-llama/create-llama/issues/244 - how about using Bastians suggestion?

leehuwuj: Makes sense, Python then should have the same behavior with TS.

Issue #270, opened 2024-09-02
backend: config/loaders.yaml


lambdaofgod:
When I run `poetry run generate` I get

```FileNotFoundError: [Errno 2] No such file or directory: 'config/loaders.yaml'```

Where can I find information on this file?

marcusschiesser:
`config/loaders.yaml` is usually generated by create-llama - what configuration did you use?

lambdaofgod:
I created the project without any example data.

leehuwuj:
the `loader.yml` config file is generated dynamically depending on the data source you pick. If you select no data source then you don't need to run `poetry run generate`

PR #279, opened 2024-09-09
chore: Always use file loader as default loader

leehuwuj:
For better DX, fix: https://github.com/run-llama/create-llama/issues/270

### DIFF
--- a/helpers/index.ts
+++ b/helpers/index.ts
@@ -96,10 +96,11 @@ async function generateContextData(
   }
 }
 
-const copyContextData = async (
+const prepareContextData = async (
   root: string,
   dataSources: TemplateDataSource[],
 ) => {
+  await makeDir(path.join(root, "data"));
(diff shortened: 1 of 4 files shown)


helpers/index.ts
```diff
@@ -174,6 +174,9 @@ export const installTemplate = async (
       await createBackendEnvFile(props.root, props);
     }
 
+    // Always create a data directory
+    await makeDir(path.join(props.root, "data"));
```
marcusschiesser, line 178: shouldn't this be at the place where we copy the data folder?

leehuwuj: Thanks, i just moved it to copy data folder:
https://github.com/run-llama/create-llama/pull/279/commits/44e193168bbf41e1a9753d2e8b6dc46cb1034550

marcusschiesser (approved):

PR #336, opened 2024-10-03
feat: add task selector to TS multiagent and revise the prompt

leehuwuj:


templates/components/multiagent/typescript/workflow/agents.ts
```diff
@@ -44,7 +44,6 @@ export const createWriter = (chatHistory: ChatMessage[]) => {
 You are given the task of writing a blog post based on research content provided by the researcher agent. Do not invent any information yourself. 
 It's important to read the entire conversation history to write the blog post accurately.
 If you receive a review from the reviewer, update the post according to the feedback and return the new post content.
-If the user requests an update with new information but no research content is provided, you must respond with: "I don't have any research content to write about."
```
leehuwuj, line 47: This prompt is useful for cases where there is an error with the tool, but it also makes the other agent think that there is no research content at all. We should better handle the tool errors another way, so I just remove this. It doesn't hurt much.

templates/components/multiagent/python/app/examples/workflow.py
```diff
@@ -171,7 +180,10 @@ async def write(
         if ev.is_good or too_many_attempts:
             # too many attempts or the blog post is good - stream final response if requested
             result = await self.run_agent(
-                ctx, writer, ev.input, streaming=ctx.data["streaming"]
+                ctx,
+                writer,
+                f"Trim the review from the reviewer and return only the blog post content. Here is the content: {ev.input}",
```
leehuwuj, line 185: Better to not show the review message from reviewer to the user.

templates/components/multiagent/python/app/examples/workflow.py
```diff
@@ -128,10 +127,20 @@ async def _decide_workflow(
         self, input: str, chat_history: List[ChatMessage]
     ) -> str:
         prompt_template = PromptTemplate(
-            "Given the following chat history and new task, decide whether to publish based on existing information.\n"
-            "Chat history:\n{chat_history}\n"
-            "New task: {input}\n"
-            "Decision (respond with either 'not_publish' or 'publish'):"
+            dedent("""
+                You are an expert in decision-making, helping people write and publish blog posts.
```
leehuwuj, line 131: Small tweak to improve the context for the LLM.

templates/components/multiagent/python/app/examples/workflow.py
```diff
@@ -171,7 +180,10 @@ async def write(
         if ev.is_good or too_many_attempts:
             # too many attempts or the blog post is good - stream final response if requested
             result = await self.run_agent(
-                ctx, writer, ev.input, streaming=ctx.data["streaming"]
+                ctx,
+                writer,
+                f"Trim the review from the reviewer and return only the blog post content. Here is the content: {ev.input}",
+                streaming=ctx.data["streaming"],
```
marcusschiesser, line 186: don't we store the generated content in `ctx.data["result"]`?
```suggestion
                f"Return only the blog post content. Here is the content: {ctx.data["result"]}",
                streaming=ctx.data["streaming"],
```

leehuwuj: It's easier to follow the workflow with `ctx.data["result"]`, but using `ev.input` is a good choice too since it includes the reviewer's suggestions for improving the post, making the final response even better.

marcusschiesser: fair point as this is the writer - but then the prompt should be different (not just removing the review but applying the reviewers suggestions)

leehuwuj: >not just removing the review but applying the reviewers suggestions 

Good idea

marcusschiesser (approved):

PR #402, opened 2024-10-28
integration: Add Hugging Face local models; ST for embeddings

tomaarsen:
Hello!

## Pull Request overview
* Add Hugging Face local models (both LLMs & embeddings)
* Use Sentence Transformers (with ONNX) for local embeddings

## Details
I did some simple testing locally, which worked. I've also added some environment variables for the embedding model, for example to specify the `backend` as torch, ONNX or OpenVINO. Sentence Transformers with ONNX is equivalent to FastEmbed, except the latter requires that the model was already pre-exported. This means they only support ~6 models, whereas Sentence Transformers supports almost 10k: https://huggingface.co/models?library=sentence-transformers.

cc @marcusschiesser as we discussed this on LinkedIn.

- Tom Aarsen

## Release Notes

- **New Features**
	- Introduced support for Hugging Face models, enhancing model compatibility and performance.
	- Added new environment variables for configuring Sentence Transformers.
	- New initialization functions for Hugging Face models in settings.
	- Interactive model selection for Hugging Face through user prompts.

- **Bug Fixes**
	- Updated dependency management to replace outdated libraries with Hugging Face dependencies.

- **Documentation**
	- Expanded `ModelProvider` options to include "huggingface."

These changes collectively improve the flexibility and functionality of the application for users leveraging AI model integrations.


helpers/providers/huggingface.ts
```diff
@@ -0,0 +1,61 @@
...
+import { ModelConfigParams } from ".";
+import { questionHandlers, toChoice } from "../../questions/utils";
+
+const MODELS = ["HuggingFaceH4/zephyr-7b-alpha"];
+type ModelData = {
+  dimensions: number;
+};
+const EMBEDDING_MODELS: Record<string, ModelData> = {
+  "all-MiniLM-L6-v2": { dimensions: 384 },
+};
```
marcusschiesser, line 11: @tomaarsen please add other models that are running well locally. (I added a model selector that you can call with `create-llama --ask-models`)

templates/components/settings/python/settings.py
```diff
@@ -168,8 +181,8 @@ def init_anthropic():
     }
 
     Settings.llm = Anthropic(model=model_map[os.getenv("MODEL")])
-    # Anthropic does not provide embeddings, so we use FastEmbed instead
-    init_fastembed()
+    # Anthropic does not provide embeddings, so we use open Sentence Transformer models instead
```
marcusschiesser, line 184: @tomaarsen we actually had a PR that was replacing sentence transformers (`llama-index-embeddings-huggingface@0.2.0`) with fastembed (see https://github.com/run-llama/create-llama/pull/162/files) - the reasons have been:
1. sentence transformers was a too large dependency
2. we had problems running it in the docker container of RAGapp (some pytorch issue that I forgot about)
before we replace it back - how is the situation now?

marcusschiesser (commented):
@tomaarsen thanks a lot for the PR - I added support for the askModels feature and please have a look at my comments

tomaarsen:
Hello @marcusschiesser,

Apologies for my delay, I had some issues on my projects that I needed to tackle first. 

I saw #162, although I didn't know the reasoning other than "improved embedding model usage". As for the dependency size - the size is primarily due to the `torch` dependency, which installs with CUDA by default on Linux, even if your device doesn't have a GPU. In that case, it's preferable to first install `torch` with `pip install torch --index-url https://download.pytorch.org/whl/cpu`.

If you do have a GPU & you'd like to run locally, you're usually best off accepting the large `torch` dependency.

As for the docker issue that you described: I'm not aware of any `torch`-specific docker issue, in theory there shouldn't be any incompatibilities with docker and `torch`.

Lastly, I'll make a follow-up to #414 to add more recommended embedding models.

- Tom Aarsen

PR #449, opened 2024-11-25
chore: bump ai v4

thucpn:
## Summary by CodeRabbit

- **New Features**
	- Updated AI dependency to version 4.0.3, potentially introducing new features and improvements.
	- Enhanced chat response handling with a new method for streaming data responses.

- **Bug Fixes**
	- Improved error handling in chat response functions to ensure consistent responses during exceptions.

- **Chores**
	- Updated package.json files to reflect new dependency versions across multiple projects.

### DIFF
--- a/templates/components/multiagent/typescript/nextjs/route.ts
+++ b/templates/components/multiagent/typescript/nextjs/route.ts
@@ -1,5 +1,6 @@
 import { initObservability } from "@/app/observability";
-import { StreamingTextResponse, type Message } from "ai";
+import { LlamaIndexAdapter, parseDataStreamPart, type Message } from "ai";
+import { EngineResponse } from "llamaindex";
 import { NextRequest, NextResponse } from "next/server";
 import { initSettings } from "./engine/settings";
 import {
@@ -39,11 +40,10 @@ export async function POST(request: NextRequest) {
       message: userMessageContent,
       streaming: true,
     });
-    const { stream, dataStream } =
+    const { stream, dataStream: data } =
       await createStreamFromWorkflowContext(context);
-
-    // Return the two streams in one response
-    return new StreamingTextResponse(stream, {}, dataStream);
+    const streamIterable = streamToAsyncIterable(stream);
+    return LlamaIndexAdapter.toDataStreamResponse(streamIterable, { data });
   } catch (error) {
     console.error("[LlamaIndex]", error);
     return NextResponse.json(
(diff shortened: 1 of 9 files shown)


templates/types/streaming/express/src/controllers/chat.controller.ts
```diff
@@ -59,8 +59,23 @@ export const chat = async (req: Request, res: Response) => {
...
+      reader?.read().then(({ done, value }: { done: boolean; value?: any }) => {
+        if (done) {
+          res.end();
+          return;
+        }
+        res.write(value);
+        read();
+      });
+    }
+    read();
```
marcusschiesser, line 78: why not using await?
```suggestion
    if (response.body) {
      const reader = response.body.getReader();
      while (true) {
        const { done, value } = await reader.read();
        if (done) {
          res.end();
          return;
        }
        res.write(value);
      }
    }
```

templates/components/multiagent/typescript/nextjs/route.ts
```diff
@@ -41,9 +44,10 @@ export async function POST(request: NextRequest) {
     });
     const { stream, dataStream } =
       await createStreamFromWorkflowContext(context);
-
-    // Return the two streams in one response
-    return new StreamingTextResponse(stream, {}, dataStream);
+    const streamIterable = streamToAsyncIterable(stream);
+    return LlamaIndexAdapter.toDataStreamResponse(streamIterable, {
+      data: dataStream,
+    });
```
marcusschiesser, line 50: how about calling `streamToAsyncIterable` inside of `createStreamFromWorkflowContext`
```suggestion
    return LlamaIndexAdapter.toDataStreamResponse(stream, {
      data: dataStream,
    });
```

templates/components/multiagent/typescript/workflow/stream.ts
```diff
@@ -46,15 +45,16 @@ export async function createStreamFromWorkflowContext<Input, Output, Context>(
         closeStreams(controller);
         return;
       }
-      const text = trimStartOfStream(chunk.delta ?? "");
-      if (text) {
-        controller.enqueue(encoder.encode(text));
+      if (chunk.delta) {
+        controller.enqueue(encoder.encode(chunk.delta));
```
marcusschiesser, line 49: can't you directly send `EngineResponse` objects here? then you don't need to do `streamToAsyncIterable`

templates/components/multiagent/typescript/workflow/stream.ts
```diff
@@ -46,15 +45,16 @@ export async function createStreamFromWorkflowContext<Input, Output, Context>(
...
       }
     },
   });
 
   return {
-    stream: mainStream.pipeThrough(createStreamDataTransformer()),
+    stream: mainStream
+      .pipeThrough(createStreamDataTransformer())
+      .pipeThrough(new TextDecoderStream()),
```
marcusschiesser, line 57: if you use `LlamaIndexAdapter`, it takes care of piping dataStream through the text stream
```suggestion
    stream: mainStream
```

PR #456, opened 2024-12-03
Bump llamacloud index and fix issues

leehuwuj:
## Summary by CodeRabbit

- **New Features**
	- Enhanced index management with new functions for creating and retrieving indices.
	- Improved error handling during file uploads with user notifications.

- **Bug Fixes**
	- Resolved breaking changes in the LlamaCloud library to ensure smooth operation.

- **Chores**
	- Updated various dependency versions across tools and libraries for better performance and features.
	- Improved clarity and consistency in code structure and logging.
	- Adjusted UI component sizes for better visibility.

### DIFF
--- a/templates/components/vectordbs/python/llamacloud/index.py
+++ b/templates/components/vectordbs/python/llamacloud/index.py
@@ -95,10 +95,9 @@ def get_index(
         index = LlamaCloudIndex(**config.to_index_kwargs())
         return index
     except ValueError:
+        logger.warning("Index not found")
         if create_if_missing:
-            logger.info(
-                f"Index {config.llama_cloud_pipeline_config.pipeline} not found, creating it"
-            )
+            logger.info("Creating index")
             _create_index(config)
             return LlamaCloudIndex(**config.to_index_kwargs())
         return None
--- a/templates/components/vectordbs/python/llamacloud/index.py
+++ b/templates/components/vectordbs/python/llamacloud/index.py
@@ -2,10 +2,12 @@ import logging
 import os
 from typing import Optional
 
+from llama_cloud import PipelineType
 from llama_index.core.callbacks import CallbackManager
 from llama_index.core.ingestion.api_utils import (
     get_client as llama_cloud_get_client,
 )
+from llama_index.core.settings import Settings
 from llama_index.indices.managed.llama_cloud import LlamaCloudIndex
 from pydantic import BaseModel, Field, field_validator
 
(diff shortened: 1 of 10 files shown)


templates/components/vectordbs/python/llamacloud/index.py
```diff
@@ -82,14 +84,63 @@ def to_index_kwargs(self) -> dict:
...
 
 
-def get_index(config: IndexConfig = None):
+def get_index(
+    config: IndexConfig = None,
+    create_if_missing: bool = False,
+):
     if config is None:
         config = IndexConfig()
-    index = LlamaCloudIndex(**config.to_index_kwargs())
```
leehuwuj, line 88: The newly update resolve and throw the index if it's not exist automatically, see: https://github.com/run-llama/llama_index/commit/a56c3e0b78adf224f084e3f8e6b43408e8281c0c#diff-52ef3b6a93ae4dc22f541ee65207a1f7a22b4724ec0f757e3b7f221dd480b2b6R90

templates/components/vectordbs/python/llamacloud/index.py
```diff
@@ -82,14 +84,63 @@ def to_index_kwargs(self) -> dict:
...
-    return index
+    # Check whether the index exists
+    try:
+        index = LlamaCloudIndex(**config.to_index_kwargs())
+        return index
+    except ValueError:
+        logger.warning("Index not found")
+        if create_if_missing:
+            logger.info("Creating index")
+            _create_index(config)
```
leehuwuj, line 101: We don't need the ensure function because it's now resolved by the framework. Instead, i moved the create index code to index file and call it here

templates/components/vectordbs/python/llamacloud/service.py
```diff
@@ -79,12 +80,14 @@ def add_file_to_pipeline(
         max_attempts = 20
         attempt = 0
         while attempt < max_attempts:
-            result = client.pipelines.get_pipeline_file_status(pipeline_id, file.id)
+            result = client.pipelines.get_pipeline_file_status(
+                file_id=file_id, pipeline_id=pipeline_id
```
leehuwuj, line 84: The get_pipeline_file_status() function has changed the ordering of the parameters. It is better to use keyword arguments.

templates/components/vectordbs/python/llamacloud/service.py
```diff
@@ -67,10 +67,11 @@ def add_file_to_pipeline(
     ) -> str:
         client = get_client()
         file = client.files.upload_file(project_id=project_id, upload_file=upload_file)
+        file_id = file.id
```
marcusschiesser, line 70: why this line?

templates/components/vectordbs/python/llamacloud/service.py
marcusschiesser: we should add a parameter to not wait for the file being processed. then from `generate.py` we don't wait and when the user is uploading a file, we wait

templates/components/vectordbs/python/llamacloud/service.py
leehuwuj: It is used in a few places later, but llama-index frequently changes how to access the ID, so I think it is better to declare it once here for easier editing or to avoid missing changes, and it looks to me that our current mypy configuration didn't check that.

templates/components/vectordbs/python/llamacloud/service.py
leehuwuj: Thanks! I just remembered this.

PR #482, opened 2025-01-17
feat: Add deep research use case (Python)

leehuwuj:
## Release Notes

- **New Features**
  - Introduced deep research workflow for analyzing personal documents
  - Added support for conducting in-depth research using multi-agent system
  - Enhanced chat interface with new deep research event visualization

- **Improvements**
  - Updated dependency versions for better compatibility
  - Expanded template use cases and application types
  - Improved event streaming and background task handling

- **Documentation**
  - Added README template for deep research project setup
  - Included detailed workflow and configuration instructions


templates/components/agents/python/blog/app/workflows/models.py
```diff
@@ -0,0 +1,63 @@
...
+    question: str
+    answer: str
+
+
+class WriteReportEvent(Event):
+    pass
+
+
+## Stream events
+class ResponseChunkEvent(Event):
```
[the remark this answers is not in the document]
leehuwuj: As we are now always stream the text at the end of workflow so the stream combine is not needed: https://github.com/run-llama/create-llama/blob/eec237c5feea1af9cdd5b276d34ebe3b8d0fd185/templates/components/multiagent/python/app/api/routers/vercel_response.py#L100

templates/types/streaming/nextjs/app/components/ui/chat/chat-message-content.tsx
```diff
@@ -22,6 +23,11 @@ export function ChatMessageContent() {
         />
       ),
     },
+    // add the writer card
```
leehuwuj, line 26: @thucpn @marcusschiesser 
I always include WriterCard as a custom component for all use cases. Do you have any better ideas?

templates/components/agents/python/blog/app/workflows/agents.py
```diff
@@ -0,0 +1,151 @@
...
+
+async def research(
+    question: str,
+    context_nodes: List[Node],
+) -> str:
+    prompt = """
+    You are a researcher who are inprogress of answering the question.
+    The purpose is to answer the question based on the collected information, no prior knowledge or making up any new information.
+    Always add citations to the sentence/point/paragraph to the id of the provided content.
+    The citation should follow this format: [citation:id]() where id is the id of the content.
```
leehuwuj, line 77: This is for citation. I think that using the node_id is a good idea as it doesn't change during the workflow - and the citation feature in RAG template also follows this way. The difference is that, in the workflow, I'm using a retriever instead of a query engine, which I'll include the node_id to the LLM content myself while RAG template use node post processor.

templates/types/streaming/nextjs/app/components/ui/chat/custom/writer-card.tsx
```diff
@@ -0,0 +1,193 @@
...
+    state: EventState | null;
+  };
+  analyze: {
+    state: EventState | null;
+    questions: QuestionState[];
+  };
+};
+
+// Update the state based on the event
+const updateState = (state: WriterState, event: WriterEvent): WriterState => {
```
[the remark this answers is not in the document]
leehuwuj: Sorry, it's just a pure transform function. I updated the name to avoid confusion

helpers/python.ts
```diff
@@ -480,6 +480,13 @@ export const installPythonTemplate = async ({
     await copyRouterCode(root, tools ?? []);
   }
 
+  // Copy multiagents overrides
+  if (template === "multiagent") {
+    await copy("**", path.join(root), {
+      cwd: path.join(compPath, "multiagent", "python"),
+    });
+  }
```
leehuwuj, line 488: somehow we forgot to copy overrides code for multi-agent; it fallback to chat engine so we didn't have any e2e error

templates/types/streaming/nextjs/app/components/ui/chat/custom/writer-card.tsx
```diff
@@ -0,0 +1,193 @@
...
+          state: event.state,
+        },
+      };
+
+    default:
+      return state;
+  }
+};
+
+export function WriterCard({ message }: { message: Message }) {
```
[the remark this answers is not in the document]
leehuwuj: The other chat message components also pass `message` as props.

templates/types/streaming/nextjs/app/components/ui/chat/chat-message-content.tsx
leehuwuj: It's a bit weird that we add a component for a use case to the chat UI. At the moment, I don't think we can make the writer card more generic, so I'll just leave it here. I think it would be great if we could enable the general events/components in the chat UI and also have an easy way to add custom components to the front end.

templates/components/multiagent/python/app/workflows/events.py
```diff
@@ -25,3 +28,18 @@ def to_response(self) -> dict:
                 "data": self.data,
             },
         }
+
+
+class SourceNodesEvent(Event):
```
marcusschiesser, line 33: we can probably also move this to the workflow

templates/types/streaming/nextjs/app/components/ui/chat/custom/writer-card.tsx
```diff
@@ -0,0 +1,201 @@
...
+  );
+
+  const state = useMemo(() => writeEventsToState(writerEvents), [writerEvents]);
+
+  if (!hasWriterEvents) {
+    return null;
+  }
+
+  return (
+    <div className="bg-white rounded-2xl shadow-xl p-5 space-y-6 text-gray-800 w-full">
```
marcusschiesser, line 147: maybe too much CSS? better use shadcn style @thucpn what do you think?

templates/components/multiagent/python/app/workflows/events.py
leehuwuj: I think that the SourceNodesEvent would be helpful for other workflows later. We can reuse it to display the retrieved nodes from the workflow.

leehuwuj: Ah, i just read your concern here: https://github.com/run-llama/create-llama/pull/482#discussion_r1923113395
Already thought about that, but afterall i think we should keep the AgentRunEvent and SourceNodesEvent in a shared code for reusing.
The current is `app/workflows` which is the same directory as the workflow code. How about we move it into a sub-folder? Like `app/workflows/agent` or `app/workflows/shared` or `app/workflows/common`?

PR #495, opened 2025-01-24
Standardize the code of workflow use cases

leehuwuj:
## Summary by CodeRabbit

- **Refactor**
  - Updated workflows to capture conversation history dynamically at runtime rather than via static initialization, streamlining interaction handling.

- **New Features**
  - Enhanced streaming responses with improved dynamic control, error handling, and cancellation processes for a more robust and responsive user experience.
  - Introduced a new `stream` attribute for better management of streaming logic across workflows.
  - Added a new entry for "create-llama" to standardize workflow use cases in Python.
  - Introduced new classes for handling event callbacks, including `EventCallback`, `LlamaCloudFileDownload`, `SuggestNextQuestions`, and `StreamHandler`.
  - Improved handling of user messages and chat history within workflows, allowing for more flexible interactions.

### DIFF
--- a/templates/components/engines/python/agent/engine.py
+++ b/templates/components/engines/python/agent/engine.py
@@ -1,9 +1,7 @@
 import os
 from typing import List
 
-from llama_index.core.agent import AgentRunner
-from llama_index.core.callbacks import CallbackManager
-from llama_index.core.settings import Settings
+from llama_index.core.agent.workflow import AgentWorkflow
 from llama_index.core.tools import BaseTool
 
 from app.engine.index import IndexConfig, get_index
@@ -11,13 +9,14 @@ from app.engine.tools import ToolFactory
 from app.engine.tools.query_engine import get_query_engine_tool
 
 
-def get_chat_engine(params=None, event_handlers=None, **kwargs):
+def get_chat_engine(params=None, **kwargs) -> AgentWorkflow:
     system_prompt = os.getenv("SYSTEM_PROMPT")
     tools: List[BaseTool] = []
-    callback_manager = CallbackManager(handlers=event_handlers or [])
+    if params is None:
+        params = {}
 
     # Add query tool if index exists
-    index_config = IndexConfig(callback_manager=callback_manager, **(params or {}))
+    index_config = IndexConfig(**params)
     index = get_index(index_config)
     if index is not None:
         query_engine_tool = get_query_engine_tool(index, **kwargs)
@@ -27,10 +26,7 @@ def get_chat_engine(params=None, event_handlers=None, **kwargs):
     configured_tools: List[BaseTool] = ToolFactory.from_env()
     tools.extend(configured_tools)
 
-    return AgentRunner.from_llm(
-        llm=Settings.llm,
-        tools=tools,
+    return AgentWorkflow.from_tools_or_functions(
+        tools_or_functions=tools,  # type: ignore
         system_prompt=system_prompt,
-        callback_manager=callback_manager,
-        verbose=True,
     )
--- a/templates/components/engines/python/agent/engine.py
+++ b/templates/components/engines/python/agent/engine.py
@@ -1,7 +1,9 @@
 import os
 from typing import List
 
-from llama_index.core.agent.workflow import AgentWorkflow
+from llama_index.core.agent import AgentRunner
+from llama_index.core.callbacks import CallbackManager
+from llama_index.core.settings import Settings
 from llama_index.core.tools import BaseTool
 
 from app.engine.index import IndexConfig, get_index
@@ -9,14 +11,13 @@ from app.engine.tools import ToolFactory
 from app.engine.tools.query_engine import get_query_engine_tool
 
 
-def get_chat_engine(params=None, **kwargs) -> AgentWorkflow:
+def get_chat_engine(params=None, event_handlers=None, **kwargs):
     system_prompt = os.getenv("SYSTEM_PROMPT")
     tools: List[BaseTool] = []
-    if params is None:
-        params = {}
+    callback_manager = CallbackManager(handlers=event_handlers or [])
 
     # Add query tool if index exists
-    index_config = IndexConfig(**params)
+    index_config = IndexConfig(callback_manager=callback_manager, **(params or {}))
     index = get_index(index_config)
     if index is not None:
         query_engine_tool = get_query_engine_tool(index, **kwargs)
@@ -26,7 +27,10 @@ def get_chat_engine(params=None, **kwargs) -> AgentWorkflow:
     configured_tools: List[BaseTool] = ToolFactory.from_env()
     tools.extend(configured_tools)
 
-    return AgentWorkflow.from_tools_or_functions(
-        tools_or_functions=tools,  # type: ignore
+    return AgentRunner.from_llm(
+        llm=Settings.llm,
+        tools=tools,
         system_prompt=system_prompt,
+        callback_manager=callback_manager,
+        verbose=True,
     )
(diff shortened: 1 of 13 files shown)


templates/components/multiagent/python/app/api/routers/vercel_response.py
```diff
@@ -38,24 +42,50 @@ def __init__(
...
+                        yield self.convert_data(
+                            {
+                                "type": "events",
+                                "data": {
+                                    "title": f"Calling tool: {event.tool_name} with input: {event.tool_kwargs['input']}",
+                                },
+                            }
+                        )
+                if isinstance(event, AgentStream):
+                    yield self.convert_text(event.delta)
```
leehuwuj, line 75: We should stream the response from an agent event; let's use AgentStream from the framework! 
TODO: Update the current workflow use cases.

templates/components/multiagent/python/app/api/routers/vercel_response.py
```diff
@@ -38,24 +42,50 @@ def __init__(
...
-                async for output in streamer:
-                    if not is_stream_started:
-                        is_stream_started = True
-                        # Stream a blank message to start the stream
-                        yield self.convert_text("")
-
-                    yield output
+            async for event in events:
+                if isinstance(event, ToolCall):
+                    if hasattr(event, "tool_output"):
```
leehuwuj, line 48: For now, i need to transform the workflow event into the UI event schema. But it'd be great if we just use the original schema in the UI because we now have the standard event from the framework.
We can also consider to apply that approach for our own customized events (without `to_response()` method)

templates/components/multiagent/python/app/api/routers/vercel_response.py
```diff
@@ -38,24 +42,50 @@ def __init__(
...
+                                },
+                            }
+                        )
+                if isinstance(event, AgentStream):
+                    yield self.convert_text(event.delta)
+                if hasattr(event, "to_response"):
+                    event_response = event.to_response()
+                    if event_response is not None:
+                        yield self.convert_data(event_response)
+                    # TODO: Add a standard way to process nodes
```
leehuwuj, line 80: We will have an issue with source nodes feature.
1. The basic AgentWorkflow doesn't have an event for source nodes. We must extract the nodes from the tool call result event.
2. In the other hand, a customized workflow (e.g deep research) can manually extract and yield a event specific for source nodes.

At the moment, i think we need to handle both 1. and 2. but it'd be great if we standardize it into a single approach.

templates/components/engines/python/agent/engine.py
leehuwuj, line 1: TBD: Where should we put this file?

1. In the `engine` folder
2. In the `workflows` folder

templates/components/multiagent/python/app/api/routers/vercel_response.py
marcusschiesser: yes better forward events to the ui 1:1 and let the UI deal with events (like we do with deep research)

templates/components/multiagent/python/app/api/routers/vercel_response.py
marcusschiesser: the more framework the better :)

templates/components/multiagent/python/app/api/routers/vercel_response.py
marcusschiesser: how about putting this concern in a special adapter class?

templates/components/multiagent/python/app/api/routers/vercel_response.py
```diff
@@ -26,135 +24,75 @@ class VercelStreamResponse(StreamingResponse):
...
-                "LlamaCloud is not configured. Skipping post processing of nodes"
-            )
-            pass
+            logger.error(f"Error in stream response: {e}")
+            yield self.convert_error(str(e))
+            await self.handler.cancel_run()
+
+    async def _stream_text(
+        self, event: AgentStream | StopEvent
+    ) -> AsyncGenerator[str, None]:
```
leehuwuj, line 65: I think we can accept streaming text from both AgentStream (for multi-agent pattern) and StopEvent (workflow).

templates/components/multiagent/python/app/api/routers/vercel_response.py
```diff
@@ -26,135 +24,75 @@ class VercelStreamResponse(StreamingResponse):
...
+        """
+        Accept stream text from either AgentStream or StopEvent with string or AsyncGenerator result
+        """
+        if isinstance(event, AgentStream):
+            yield self.convert_text(event.delta)
+        elif isinstance(event, StopEvent):
+            if isinstance(event.result, str):
```
leehuwuj, line 72: There is the case that we need to send a manual response from the workflow which only have one text chunk.

templates/components/multiagent/python/app/workflows/tools.py
```diff
@@ -51,7 +52,9 @@ async def full_response(self) -> str:
         assert self.generator is not None
         full_response = ""
         async for chunk in self.generator:
-            full_response += chunk.message.content
+            content = chunk.message.content
```
leehuwuj, line 55: Not related to the PR, just fix an error in the case message content is None

templates/components/multiagent/python/app/api/routers/chat.py
```diff
@@ -39,10 +42,13 @@ async def chat(
             stream=True,
         )
         return VercelStreamResponse(
```
leehuwuj, line 44: It might be even better if we consider the vercelstream response is a part of the stream handler so we can just simply run with Vercel's FE integration.
```python
StreamHandler.from_default(....)  
return stream_handler.vercel_stream()
```

templates/components/multiagent/python/app/api/routers/chat.py
```diff
@@ -39,10 +42,13 @@ async def chat(
...
-            request=request,
-            chat_data=data,
```
marcusschiesser, line 40: what happened with request and chat_data?

templates/components/multiagent/python/app/api/routers/chat.py
marcusschiesser: wouldn't it be easier if VercelStreamResponse could directly handle callbacks, like?

```python
return VercelStreamResponse( 
  handler=handler,
  callbacks=[
                    LlamaCloudFileDownload.from_default(background_tasks),
                    SuggestNextQuestions.from_default(data),
  ]
)
```

otherwise this is also fine:
```python
return StreamHandler.from_default( 
  handler=handler,
  callbacks=[
                    LlamaCloudFileDownload.from_default(background_tasks),
                    SuggestNextQuestions.from_default(data),
  ]
).vercel_stream()
```

templates/components/multiagent/python/app/api/routers/chat.py
leehuwuj: - We use request to detect connection breaking: https://github.com/run-llama/create-llama/blob/c1552ebb00f3b531101ef500c2f1cc823b3c50ac/templates/types/streaming/fastapi/app/api/routers/vercel_response.py#L60 but in the implementation for workflow we don't need that anymore. It's now still able to stop the workflow when user cancel the request.
- Chat data is passed to SuggestNextQuestions already.

templates/components/multiagent/python/app/api/routers/chat.py
leehuwuj: I updated to 
```
return StreamHandler.from_default( 
  handler=handler,
  callbacks=[
                    LlamaCloudFileDownload.from_default(background_tasks),
                    SuggestNextQuestions.from_default(data),
  ]
).vercel_stream()
```

It looks better to me :D

PR #530, opened 2025-03-28
feat: add llamacloud to llama_index_server

leehuwuj:


llama-index-server/llama_index/server/services/llamacloud/__init__.py
leehuwuj, line 1: The idea is moving the LlamaCloud code to the server, then we can simplify the logic in the generated code. Also, it's always included in the server package, we can also eliminate the ugly code that checks if the LlamaCloud service is importable in the old code.

marcusschiesser (approved):

llama-index-server/README.md
```diff
@@ -32,24 +39,91 @@ def create_workflow() -> Workflow:
...
+- `api_prefix`: API route prefix (default: "/api")
+- `ui_path`: Path for downloaded UI static files (default: ".ui")
+
+## Default Routers and Features
+
+### Chat Router
+
+The server includes a default chat router at `/api/chat` for handling chat interactions.
+
+### Static File Serving
```
marcusschiesser, line 89: @leehuwuj as you see this doesn't make a lot of sense as our folder mapping is workflow specific.
Suggestion: we add a `file_server` parameter that we set to `['data', 'output']` for our example workflows. if it's empty there are no `api/files` routes
Do we still need then the `use_default_routers` parameter?

llama-index-server/README.md
```diff
@@ -86,12 +87,6 @@ The LlamaIndexServer accepts the following configuration parameters:
 
 The server includes a default chat router at `/api/chat` for handling chat interactions.
 
-### Static File Serving
```
marcusschiesser: @leehuwuj why did you remove this?

llama-index-server/llama_index/server/tools/interpreter.py
```diff
@@ -34,34 +33,24 @@ class E2BCodeInterpreter:
...
-        api_key: Optional[str] = None,
-        filesever_url_prefix: Optional[str] = None,
+        api_key: str,
         output_dir: Optional[str] = None,
         uploaded_files_dir: Optional[str] = None,
     ):
         """
         Args:
-            api_key: The API key for the E2B Code Interpreter. If not provided, it will be read from the environment variable `E2B_API_KEY`.
-            filesever_url_prefix: The prefix for the file server or loaded from env: `FILESERVER_URL_PREFIX`, default is `/api/files`.
```
marcusschiesser, line 45: filesever_url_prefix not needing anymore?

llama-index-server/README.md
```diff
@@ -78,6 +78,7 @@ The LlamaIndexServer accepts the following configuration parameters:
 - `starter_questions`: List of starter questions for the chat UI
 - `verbose`: Enable verbose logging
 - `api_prefix`: API route prefix (default: "/api")
+- `server_url`: The deployment URL of the server.
```
marcusschiesser, line 81: is this optional? what's the default value?

llama-index-server/llama_index/server/tools/interpreter.py
leehuwuj: Yes, the tool will call FileService, which uses the fileserver_url_prefix.

llama-index-server/README.md
leehuwuj: I thought it was confusing to you. I just added it back

llama-index-server/README.md
leehuwuj: Updated: `server_url`: The deployment URL of the server (default is None)

PR #537, opened 2025-04-03
chore: simplify local index code

leehuwuj:
## Summary by CodeRabbit

- **Refactor**
  - Simplified storage configuration by using a constant for storage directory instead of a configuration class.
  - Enhanced workflow initialization with improved error handling for index validation before retrieving the query tool.
- **Chores**
  - Added an update patch that further simplifies local index management.

### DIFF
--- a/llama-index-server/llama_index/server/server.py
+++ b/llama-index-server/llama_index/server/server.py
@@ -1,23 +1,45 @@
 import json
 import logging
 import os
-from typing import Any, Callable, Optional
+from typing import Any, Callable, Optional, Union
 
 from fastapi import FastAPI
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.staticfiles import StaticFiles
+from pydantic import BaseModel
+
 from llama_index.core.workflow import Workflow
 from llama_index.server.api.routers.chat import chat_router
 from llama_index.server.chat_ui import download_chat_ui
 from llama_index.server.settings import server_settings
 
 
+class UIConfig(BaseModel):
+    enabled: bool = True
+    app_title: str = "LlamaIndex Server"
+    starter_questions: Optional[list[str]] = None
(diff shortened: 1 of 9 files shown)


llama-index-server/llama_index/server/server.py
```diff
@@ -1,32 +1,53 @@
...
 
     def __init__(
         self,
         workflow_factory: Callable[..., Workflow],
         logger: Optional[logging.Logger] = None,
         use_default_routers: Optional[bool] = True,
         env: Optional[str] = None,
-        include_ui: Optional[bool] = None,
-        starter_questions: Optional[list[str]] = None,
+        ui_config: Optional[Union[UIConfig, dict]] = None,
```
leehuwuj, line 50: @thucpn  How about we use a single parameter for the UI config?

templates/types/llamaindexserver/fastapi/app/index.py
```diff
@@ -5,28 +5,19 @@
...
 logger = logging.getLogger("uvicorn")
 
-
-class IndexConfig(BaseModel):
-    storage_dir: str = "storage"
-
-    @classmethod
-    def from_default(cls, chat_request: Optional[ChatRequest] = None) -> "IndexConfig":
-        return cls()
+STORAGE_DIR = "storage"
```
marcusschiesser, line 11: generate.py is using `os.environ.get("STORAGE_DIR", "storage")`

use just `storage` there too?

leehuwuj: great catch! i added back the usage of `os.environ.get("STORAGE_DIR", "storage")`. the reason is this is a component code for legacy template as well so it'd be different import path in generate.py file

marcusschiesser (approved):

Issue #546, opened 2025-04-08
Generate examples with UV instead of Poetry 


JasonG98:
Hello,

I would like to inquire if there are any plans to transition from using Poetry to UV as the primary package management tool?

While Poetry is an excellent tool,UV offers some unique features and advantages,such as faster installation speeds and simpler dependency management.I am curious to know if the team has considered this option or if there have been any discussions or plans regarding this transition.

If there have already been discussions,could you provide more information?If not,would it be possible to consider adding this proposal to future plans?

Thank you for your time and consideration.

marcusschiesser:
@JasonG98 great idea, added it to our roadmap

PR #578, opened 2025-04-19
Use uv as the default package manager and deprecate poetry.

leehuwuj:
Due to the different pyproject.toml standards, we deprecate support for Poetry and use UV as the default manager for Python templates to simplify the create-llama code base.
Fix: https://github.com/run-llama/create-llama/issues/546

## Summary by CodeRabbit

- **New Features**
  - Introduced support for the Astral uv tool for Python dependency management and environment setup, replacing Poetry across the project.
  - Added new helper utilities for interacting with uv.

- **Refactor**
  - Migrated all project configuration files from Poetry to the standardized PEP 621 format and switched build systems to Hatchling.
  - Updated Dockerfiles and GitHub Actions workflows to use uv for installation and running commands.
  - Replaced all CLI instructions, scripts, and error messages referencing Poetry with uv equivalents in documentation and code.
  - Improved dependency version formatting and handling for Python projects.

- **Bug Fixes**
  - Updated default values and decorators in Python components for improved compatibility.

- **Documentation**
  - Revised all README and deployment instructions to reflect the switch from Poetry to uv for setup, development, and production commands.


leehuwuj:
@coderabbitai review

templates/types/llamaindexserver/fastapi/run.py
```diff
@@ -0,0 +1,21 @@
+import os
+import subprocess
+
+
+def run(env: str):
```
marcusschiesser, line 5: can we keep this in one file? some idea to simplify this:
https://grok.com/share/bGVnYWN5_99f77d67-83ae-4ed4-b4bb-736c6815fb05

(e.g. we don't need APP_HOST and i don't see a usage of APP_ENV)
if it works with e2e tests, we can also hard code the port.

the goal is that the generated code is as simple as possible and users can just directly modify the generated code

leehuwuj: How about we don't specifiy dev and prod?
https://docs.astral.sh/uv/guides/integration/fastapi/
Then the user can just.
`development`
```shell
uv run fastapi dev
```
and `production`
```
uv run fastapi run
```

marcusschiesser: that's also good

templates/types/llamaindexserver/fastapi/pyproject.toml
```diff
@@ -1,35 +1,37 @@
...
-python = ">=3.11,<3.14"
-python-dotenv = "^1.0.0"
-pydantic = "<2.10"
-aiostream = "^0.5.2"
-llama-index-core = "^0.12.28"
-llama-index-server = "^0.1.14"
+[project.scripts]
+generate = "generate:generate_index"
+generate_index = "generate:generate_index"
+generate_ui = "generate:generate_ui_for_workflow"
```
marcusschiesser, line 28: generate:ui translates to generate_ui ?

templates/types/llamaindexserver/fastapi/main.py
```diff
@@ -14,33 +12,17 @@
...
     app = LlamaIndexServer(
         workflow_factory=create_workflow,  # A factory function that creates a new workflow for each request
         ui_config=UIConfig(
             component_dir=COMPONENT_DIR,
             app_title="Chat App",
         ),
-        env=env,
         logger=logger,
     )
-    # You can also add custom routes to the app
```
marcusschiesser, line 28: how about keep this so user know they can add custom routes?
but clarify that its FastAPI:
```You can also add custom FastAPI routes to app````

templates/types/llamaindexserver/fastapi/pyproject.toml
marcusschiesser: but you're using `generate:ui` in the docs

PR #610, opened 2025-05-12
feat: Add dev mode

leehuwuj:
## Summary by CodeRabbit

- **New Features**
  - Introduced a developer mode UI for editing and restarting server workflow files directly from the chat interface when development mode is enabled.
  - Added a toggleable panel with code editor, validation, and live server restart feedback for workflow updates.
  - Enabled developer endpoints for managing workflow files via the API in development mode.
  - Enforced that developer mode UI requires running the server in development environment.
  - Added a simple chatbot agent example with setup instructions and a custom workflow.

- **Documentation**
  - Updated main and example READMEs to describe the new developer mode UI, its configuration, and usage instructions.
  - Added setup and usage guides for a simple chatbot agent example.

- **Chores**
  - Upgraded the `@llamaindex/chat-ui` dependency to version 0.4.4.


python/llama-index-server/llama_index/server/api/routers/dev.py
```diff
@@ -0,0 +1,78 @@
...
+    file_name: str
+
+class WorkflowValidationResult(BaseModel):
+    valid: bool
+    error: str
+
+def dev_router() -> APIRouter:
+    # Use a prefix here to avoid conflicts with other routers
+    # but we probably don't need to do this
+    router = APIRouter(prefix="/dev", tags=["dev"])
```
marcusschiesser, line 23: i think it's a good idea
```suggestion
    # Use a prefix here to avoid conflicts with other routers
    router = APIRouter(prefix="/dev", tags=["dev"])
```

python/llama-index-server/llama_index/server/api/routers/dev.py
```diff
@@ -0,0 +1,78 @@
...
+
+    default_workflow_file_path = "workflow.py"
```
marcusschiesser, line 25: how do we deal with the code and doc generator workflow?

python/llama-index-server/llama_index/server/server.py
```diff
@@ -100,6 +100,8 @@ def __init__(
             server_settings.set_url(server_url)
         if api_prefix:
             server_settings.set_api_prefix(api_prefix)
+        if workflow_factory:
```
marcusschiesser, line 103: when do we not have a factory?

python/llama-index-server/llama_index/server/api/routers/dev.py
leehuwuj: ideally, we should somehow use a subprocess to preload the changes - by that, we don't need to care the user code or which is the workflow file.

marcusschiesser: i think we should keep it as simple as possible, better add a "Doc workflow" and a "Code workflow" use case and allow exiting just one file

PR #616, opened 2025-05-13
feat: support dev mode for backend ts server

thucpn:


packages/server/src/handlers/files.ts
```diff
@@ -21,3 +22,49 @@ export const handleServeFiles = async (
...
+const DEFAULT_WORKFLOW_FILE_PATH = "src/app/workflow.ts"; // TODO: we can make it as a parameter in server later
+
+export const getWorkflowFile = async (
+  req: IncomingMessage,
+  res: ServerResponse,
+  filePath: string = DEFAULT_WORKFLOW_FILE_PATH,
+) => {
+  const fileExists = await promisify(fs.exists)(filePath);
+  if (!fileExists) {
+    return sendJSONResponse(res, 404, {
```
leehuwuj, line 35: We probably want to show the Dev mode is currently in beta message on loading.

packages/server/package.json
```diff
@@ -100,6 +100,7 @@
     "recharts": "^2.15.2",
     "sonner": "^2.0.3",
     "tailwind-merge": "^2.6.0",
+    "typescript": "^5.7.3",
```
marcusschiesser, line 103: that's probably we heavy dependency :D can't we use @babel/parser to validate?

marcusschiesser (approved):

PR #642, opened 2025-05-22
feat: Support citation for agentic template

leehuwuj:


python/llama-index-server/llama_index/server/api/models.py
```diff
@@ -91,6 +91,15 @@ def from_source_node(cls, source_node: NodeWithScore) -> "SourceNodes":
             url=url,
         )
 
+    @classmethod
```
marcusschiesser, line 94: can we move that llamacloud concern to the llamacloud folder?

python/llama-index-server/llama_index/server/tools/index/query.py
```diff
@@ -1,30 +1,58 @@
...
     top_k = int(os.getenv("TOP_K", 0))
     if top_k != 0 and kwargs.get("filters") is None:
         kwargs["similarity_top_k"] = top_k
 
+    if enable_citation:
+        if kwargs.get("response_synthesizer") is not None:
+            # We don't override the provided response synthesizer
+            # Just show a warning
+            logger.warning(
+                "Custom response synthesizer and citation are both used. The citation might not work as intended."
```
marcusschiesser, line 36: i guess the citation won't work at all - better throw an error?

marcusschiesser: as `Accumulate` is also not the default afaik - how about we log an info that we use Accumulate?

python/llama-index-server/pyproject.toml
```diff
@@ -61,8 +61,9 @@ dev = [
     "xhtml2pdf>=0.2.17,<1.0.0",
     "pytest-cov>=6.0.0,<7.0.0",
     "llama-cloud>=0.1.17,<1.0.0",
+    "llama-index-embeddings-openai>=0.3.1",
```
marcusschiesser, line 64: needed?

python/llama-index-server/llama_index/server/tools/index/citation_agent.py
marcusschiesser, line 1: for what do we need this file?

python/llama-index-server/llama_index/server/api/callbacks/source_nodes.py
```diff
@@ -1,29 +1,64 @@
...
+            ]  # keep query_index for backward compatibility
+        if tool_name is not None:
+            logger.warning(
+                f"Tool name has been deprecated. Please use tool_suffixes instead. "
+                f"Got tool_name: {tool_name}."
+            )
+        self.tool_name = tool_name
+        self.tool_suffixes = tool_suffixes
+
+    def is_query_engine_tool(self, tool_name: str) -> bool:
```
marcusschiesser, line 40: how can we do an instance check here - this suffix check is a bit brittle

marcusschiesser: e.g. `enable_citation` could return a new type

marcusschiesser (approved):

PR #650, opened 2025-05-26
fix: release action doesn't run custom version command

leehuwuj:
## Summary by CodeRabbit

- **Chores**
  - Updated release workflow to use a new versioning command.
  - Adjusted versioning script and downgraded package version.
  - Removed the changelog file for the server package.

- **New Features**
  - Introduced a configuration option for suggesting next questions in the server package.


.changeset/afraid-forks-taste.md
```diff
@@ -0,0 +1,5 @@
+---
```
leehuwuj, line 1: Add back the changeset and roll back to a previous version of llama_index_server (see package.json) as we haven't released it yet.

.github/workflows/release.yml
```diff
@@ -56,6 +56,8 @@ jobs:
         with:
           commit: Release ${{ steps.get-changeset-status.outputs.new-version }}
           title: Release ${{ steps.get-changeset-status.outputs.new-version }}
+          # bump versions
+          version: pnpm new-version
```
leehuwuj, line 60: Expected: A new release PR created with:
1. The changeset file is removed.
2. Both package.json and pyproject.toml are bumped to a new version.

python/llama-index-server/package.json
```diff
@@ -1,13 +1,13 @@
...
   "name": "@create-llama/llama-index-server",
   "private": true,
-  "version": "0.1.17",
+  "version": "0.1.16",
   "type": "module",
   "scripts": {
     "prebuild": "uv run -- scripts/frontend.py --mode copy",
     "build": "uv build",
     "clean": "rm -rf dist build *.egg-info",
-    "new-version": "uv run python scripts/sync_version.py && git add pyproject.toml",
```
leehuwuj, line 10: @coderabbitai Based on the changeset/action code here: https://github.com/changesets/action/blob/main/src/git.ts
I think that we don't need to run `git add` command because the `pushChanges` method would add all the changes in the working directory?

PR #658, opened 2025-05-28
feat: show document artifact after generating report

thucpn:
## Summary by CodeRabbit

- **New Features**
  - The generated document artifact is now displayed immediately after creating a report, providing instant access to the report content.
  - Reports include detailed source citations to enhance traceability and context.

- **Style**
  - Improved editor appearance by updating and adding related CSS styles.

- **Chores**
  - Upgraded UI dependencies for enhanced stability and features.
  - Updated linting configuration to streamline code quality checks.


packages/create-llama/templates/components/use-cases/python/deep_research/workflow.py
```diff
@@ -365,8 +375,44 @@ async def report(self, ctx: Context, ev: ReportEvent) -> StopEvent:
             user_request=self.user_request,
             stream=self.stream,
         )
+
+        final_response = ""
+
+        from collections.abc import AsyncGenerator
```
marcusschiesser, line 381: We should be able to have some helpers in the server package to make this easier

leehuwuj: This is incorrect
```python
from collections.abc import AsyncGenerator
```
You can just import it by `from typing import AsyncGenerator`

But i think we can simplify this, LlamaIndexServer no longer has non-streaming API so we can remove the self.stream option and always stream the final response here.

python/llama-index-server/llama_index/server/utils/stream.py
```diff
@@ -0,0 +1,45 @@
...
+        # Handle streaming response (CompletionResponseAsyncGen)
+        async for chunk in res:
+            final_response += chunk.delta or ""
+            ctx.write_event_to_stream(
+                AgentStream(
+                    delta=chunk.delta or "",
+                    response=final_response,
+                    current_agent_name=current_agent_name,
+                    tool_calls=[],
+                    raw=chunk.delta or "",
```
leehuwuj, line 38: Does it have chunk.raw?

leehuwuj: i just unsure that we should created a specific function for this? May name the function as: `writeResponseToStream` because it's updating workflow stream - not handle LLM response.

By the way, if you only need the final response but still streaming then i guess chunk.text (or maybe chunk.content) is the one (unlike in TS, python accumulates it automatically)

PR #663, opened 2025-05-28
fix: bin permission make changeset break

thucpn:


leehuwuj (approved):
LGTM! but i'm just wondering, will it go back to 644 when you run `pnpm install` on Windows?

Issue #683, opened 2025-06-09
bug: cannot catch exception raised in workflow


leehuwuj:
Can reproduce with: https://github.com/run-llama/create-llama/blob/main/python/llama-index-server/examples/private_file/custom-workflow.py

Chat something without uploading a file; the workflow raised an error, but the server only logged the error in the console without propagating it into the error event to the frontend.

PR #684, opened 2025-06-09
fix cannot catch the error raised from the workflow

leehuwuj:
fix: https://github.com/run-llama/create-llama/issues/683

## Summary by CodeRabbit

- **Bug Fixes**
  - Resolved an issue where errors raised during workflow execution could not be properly caught, improving reliability of streaming chat responses.
- **Enhancements**
  - Improved streaming response handling to ensure workflows fully complete before finishing the stream.

### DIFF
--- a/python/llama-index-server/llama_index/server/api/routers/chat.py
+++ b/python/llama-index-server/llama_index/server/api/routers/chat.py
@@ -195,6 +195,7 @@ async def _stream_content(
                 if not isinstance(event, (AgentInput, AgentSetup)):
                     yield VercelStreamResponse.convert_data(event.model_dump())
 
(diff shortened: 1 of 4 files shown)


python/llama-index-server/llama_index/server/api/routers/chat.py
```diff
@@ -195,6 +195,7 @@ async def _text_stream(
                 if not isinstance(event, (AgentInput, AgentSetup)):
                     yield VercelStreamResponse.convert_data(event.model_dump())
 
+        await handler.wait_for_completion()
```
leehuwuj, line 198: handler is a future object, we need to `await` it for error propagation.

PR #689, opened 2025-06-13
fix: stream not stop after sending HumanInputEvent

thucpn:
Summary e2e issues:

- with Python: stream not stop after sending HumanInputEvent
- with TS: no issue, but add 1 more use case caused e2e timeout. This PR enhances e2e performance for TS also


.changeset/polite-bugs-develop.md
```diff
@@ -0,0 +1,6 @@
+---
+"create-llama": patch
```
leehuwuj, line 2: Why do we need patching create-llama?

